PR #17, opened 2020-05-07
Remove kernel-headers and kernel-tools

jmflinuxtx:
These packages are subpackages of the ARK kernel in ELN, but separate
packages in Fedora.  We need to make sure that we are using the packages
from the kernel packages in ELN.

Signed-off-by: Justin M. Forbes <jforbes@fedoraproject.org>

### DIFF
--- a/eln-compose.yaml
+++ b/eln-compose.yaml
@@ -43,6 +43,8 @@ data:
 
   # Just to demonstrate stuff
   unwanted_packages:
+  - kernel-headers
+  - kernel-tools
   - xen
   - xen-devel
   - xen-hypervisor


tdawson:
I'm not sure putting kernel-headers and kernel-tools into the unwanted-packages section is going to do what you want.
And from your description, I'm not positive of what you want.
In ELN, do you want kernel-headers and kernel-tools to come from the ARK kernel?
Or, do you want kernel-headers and kernel-tools to come from their own source rpm, like Fedora?
Or, do you really not want kernel-headers and kernel-tools rpm's in RHEL9/ELN?

jmflinuxtx:
Sorry for the late reply, I am on vacation. We want kernel-headers and
kernel-tools from the ark kernel package, not from the fedora separate
packages for each.

tdawson:
That makes sense.
Having them listed on the unwanted-packages list is probrubly not going to work in the long run, because that list is for binaries, not source rpms.
But, let's leave them on there for now, so that we remember.

We'll have to figure out what to do about packages like that.  I know there is at least one other package like that, that has a different source rpm in Fedora vs RHEL.

PR #21, opened 2020-05-27
Mark btrfs-progs as unwanted

jwboyer:
Enterprise doesn't support btrfs, so there should be no need for btrfs-progs

### DIFF
--- a/configs/view-eln.yaml
+++ b/configs/view-eln.yaml
@@ -19,6 +19,10 @@ data:
   
   # Josh Boyer: Drop R from ELN to see what the fallout is.
   - R
+  
+  # Enterprise doesn't support btrfs
+  - btrfs-progs
+  - btrfs-progs-devel
 
   # Justin Forbes: These packages are subpackages of the ARK kernel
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
😭

tdawson:
Josh, is there a reason that RHEL doesn't support btrfs.  Your statement gives no explanation or reason.
Can we get a bugzilla and/or a Red Hat article about this?
My concern is that there is a proposal to have btrfs be the standard for Fedora desktop images, which might make btrfs a requirement in more places.
https://fedoraproject.org/wiki/Changes/BtrfsByDefault

sgallagher:
Red Hat Enterprise Linux doesn't support btrfs (it was in Tech Preview for RHEL 6 and 7, then dropped for 8). See https://access.redhat.com/solutions/197643

In RHEL 8 (and presumably RHEL 9), the recommended alternative is [Stratis](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_storage_devices/managing-layered-local-storage-with-stratis_managing-storage-devices)

PR #24, opened 2020-06-10
Disable unvanted Ruby related packages.

voxik:
* rubygems is provided as subpackage of ruby package.
* rubypick is not needed. It is necessary to properly work with the `%bcond_without rubypick` [[1]]. Not mentioning, that it is conditionalized just for Fedora, so it should not come into play, should it? I should probably remove the fedora condition altogether.

[1]: https://src.fedoraproject.org/rpms/ruby/blob/29816481d1e7a36950e08a1447cfae6cdda62dfb/f/ruby.spec#_71


asamalik:
@voxik It's the right place, yes!

PR #26, opened 2020-06-11
Add llvm-toolset workload

tstellar:


sst_platform_tools-llvm-toolset.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - lld-test
+  - libomp
+  - libomp-devel
+  - libomp-test
+  - compiler-rt
+  - lldb
+  - lldb-devel
+  - python3-lldb
+
+  unwanted_packages:
```
asamalik, line 33: This won't work here. You can either add it to the [view config](https://github.com/minimization/content-resolver-input/blob/master/configs/view-eln.yaml) or use the new [separate exclusion list (preferred)](https://github.com/minimization/content-resolver/blob/master/config_specs/exclusion.yaml).

tstellar:
@asamalik Ok, I've moved the unwanted packages into a separate yaml file.

asamalik:
Thanks!

PR #27, opened 2020-06-17
Add sst_cs_software_management workloads

dmach:


configs/sst_cs_software_management-unwanted.yaml
```diff
@@ -0,0 +1,43 @@
+document: feedback-pipeline-workload
```
asamalik, line 1: Can you please change this to:

```
```

The [exclusion list](https://github.com/minimization/content-resolver/blob/master/config_specs/exclusion.yaml) is a different config type.

dmach: Updated, thanks!

asamalik:
Or you know what? I can just change it for you after the merge, that makes it easier for everyone!

asamalik:
Oh, you did it already, never mind! Thanks!

PR #29, opened 2020-06-18
Add sst_front_door web console workload

martinpitt:
I have some things that are still unclear to me:

 * cockpit-podman is part of the "container-tools" module. Should that still live here, or are modules handled specially?

 * The "cockpit" package already has a few "Recommends", like cockpit-packagekit. We definitively want them in RHEL 9. Should they be repeated explicitly in the yaml here, or is it fine to rely on `Recommends:`?

Thanks!

### DIFF
--- a/configs/sst_front_door-unwanted.yaml
+++ b/configs/sst_front_door-unwanted.yaml
@@ -1,0 +1,13 @@
+document: feedback-pipeline-unwanted
+version: 1
+data:
+  name: Unwanted Packages -- sst_front_door
+  description: Packages that should not be distributed in RHEL 9
+  maintainer: sst_front_door
+  labels:
+    - cockpit
+    - web-console
+
+  unwanted_packages:
+    # only needed for integration tests (buildroot), never ship this
+    - cockpit-tests
--- a/configs/sst_front_door-web-console.yaml
+++ b/configs/sst_front_door-web-console.yaml
@@ -1,0 +1,21 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Web Console (Cockpit)
+  description: Cockpit basic packages and supported applications
+  maintainer: sst_front_door
+  labels:
+    - cockpit
+    - web-console
+
+  packages:
+    # meta-package for the basic stuff (web server, bridge, core pages)
+    - cockpit
+    # recommended packages from cockpit metapackage
+    - cockpit-packagekit
+    - cockpit-storaged
+    - cockpit-pcp
+    # web console applications developed by Front Door SST
+    - cockpit-composer
+    # in container-tools stream in RHEL 8, but not in a module in Fedora
+    - cockpit-podman


martinpitt:
@larskarlitski, @jgiardino: FYI, this includes cockpit-composer. I suppose it's fine to keep that in the same file, but if you'd rather split it out, I'm happy to do that too (or you send a PR).

asamalik:
> I have some things that are still unclear to me:
> 
> * cockpit-podman is part of the "container-tools" module. Should that still live here, or are modules handled specially?

You can enable a specific module stream in the workload definitions. See the [workload spec](https://github.com/minimization/content-resolver/blob/master/config_specs/workload.yaml#L84).

> * The "cockpit" package already has a few "Recommends", like cockpit-packagekit. We definitively want them in RHEL 9. Should they be repeated explicitly in the yaml here, or is it fine to rely on `Recommends:`?

Weak dependencies are turned off by default. You can turn them on for the whole workload, but then that applies to every single package. I'd rather recommend listing these packages, so you're explicit about what you expect there.

> 
> Thanks!

asamalik:
> Some more questions:
> 
> * This supersedes the auto-generated/imported rhel-8-*-comps files. Should this PR remove the cockpit packages from there, or are they left intact? I suppose the latter according to git log, but that may be a FAQ worth adding to the documentation.

All configs with the maintainer field set to "bakery" will be removed in the end. If there are some your SST should own, or that overlap with your workload, you can propose to remove them in a PR. Or you can claim them by setting the maintainer field, and modify as you need.

> * I saw no distinction between BaseOS and AppStream in the YAML definitions. Was I missing anything, or how is that defined in RHEL 9? (Maybe also a FAQ).

That will be later in the process. Right now we're only identifying the list. 

Good note about a FAQ, I've added it to the internal document.

> * Likewise, I saw no marking of "minimal" vs. "default" install types, what we had in comps. Is there a counterpart for this, or is it somehow not necessary any more with RHEL 9?

You can have multiple workloads representing these. See the [PR from the Software Management SST](https://github.com/minimization/content-resolver-input/pull/27/files) — they have a minimal package manager there, for example.

martinpitt:
@asamalik : Thanks for the review! I definitively want to rework this at least for the "Recommends:" and module bits.

asamalik:
All right! I've converted this PR to a draft, as we have multiple people merging these (to be responsive), so it doesn't get merged by accident. When you're done, you can just switch it back and I or someone will merge it. Cheers!

martinpitt:
THanks @asamalik ! I added the recommended packages. I did not yet enable a module stream as this isn't relevant for Fedora, only for RHEL 8, and right now it's not yet clear what the situation will be in RHEL 9. I added a comment, so supposedly at some point this will fail if the package is not present, and we can fix it  later?

I did not yet define multiple workloads yet with a "minimal" set, as it's a strict subset of the "default" one; as this is just defining the overall content set of RHEL 9, not install scenarios.

asamalik:
> THanks @asamalik ! I added the recommended packages. I did not yet enable a module stream as this isn't relevant for Fedora, only for RHEL 8, and right now it's not yet clear what the situation will be in RHEL 9. I added a comment, so supposedly at some point this will fail if the package is not present, and we can fix it later?

Yes, absolutely, it'll show as failed if the packages are no longer there.

> 
> I did not yet define multiple workloads yet with a "minimal" set, as it's a strict subset of the "default" one; as this is just defining the overall content set of RHEL 9, not install scenarios.

You're right, although, if you have any plans of perhaps minimising your minimal set, or just want to have size monitoring for free, feel free to add it, too! You'll be able to see what packages get dragged in, etc. This tool has been initially written for minimization, so it might be helpful to you as well.

martinpitt:
Thanks @asamalik for your help!

PR #22, opened 2020-06-03
Update .NET/Java ELN configuration

omajid:
I don't know what I am doing. This is me asking for advice.

1. Am I supposed to modify these files that say they are generated from rhel 8 appstream? Is there some other workload file where I should be placing things we want as candidates for ELN? Should I be creating new ones?

1. OpenJDK exists in a couple of locations:

```
configs/prototype-eln-buildroot-workload.yaml          
470:        - java-11-openjdk-devel                                                                                    
471:        - java-1.8.0-openjdk-javadoc                                                                               
472:        - java-1.8.0-openjdk-javadoc-zip
                                                           
configs/rhel-8-appstream-comps-additional-devel.yaml
26:  - java-1.8.0-openjdk-devel
                                                                                                                       
configs/rhel-8-appstream-comps-java-platform.yaml                                                                      
10:  - java-1.8.0-openjdk                                                                                              
                                                                                                                       
configs-temporarily-disabled/fedora-rawhide-comps-java-development.yaml
11:  - java-1.8.0-openjdk-devel                                                                                        
50:  - java-1.8.0-openjdk-demo                                                                                         
51:  - java-1.8.0-openjdk-javadoc                  
52:  - java-1.8.0-openjdk-src                    
                                                                                                                       
configs-temporarily-disabled/fedora-rawhide-comps-javaenterprise.yaml
10:  - java-1.8.0-openjdk          
11:  - java-1.8.0-openjdk-devel        
                                                           
configs-temporarily-disabled/fedora-rawhide-comps-java.yaml 
9:  - java-1.8.0-openjdk
```

Which one should we use if we want to manipulate ELN?

3. Some of these packages don't exist yet. We expect to introduce them to Fedora as upstream releases happen. Is that okay? 

This is mostly a guess based on:

- https://fedoraproject.org/wiki/Changes/Java11
- https://devblogs.microsoft.com/dotnet/introducing-net-5/

PR #30, opened 2020-06-19
Update jmc ELN configuration

jiekang:
Hi, this adds JMC which is also under sst_dotnet_java. I copied from what was done in https://github.com/minimization/content-resolver-input/pull/22

JMC is delivered as a module in Fedora, does that change anything?

### DIFF
--- a/configs/sst_dotnet_java-jmc.yaml
+++ b/configs/sst_dotnet_java-jmc.yaml
@@ -4,6 +4,9 @@ data:
   name: JMC
   description: JDK Mission Control application and libraries
   maintainer: sst_dotnet_java
+
+  modules_enable:
+  - jmc:latest
   
   packages: []
 
--- a/configs/sst_dotnet_java-jmc.yaml
+++ b/configs/sst_dotnet_java-jmc.yaml
@@ -1,0 +1,15 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: JMC
+  description: JDK Mission Control application and libraries
+  maintainer: sst_dotnet_java
+  
+  packages: []
+
+  arch_packages:
+    x86_64:
+    - jmc
+    - jmc-core
+  labels:
+  - eln


omajid:
@asamalik Can you help us understand how to use modules here? Is it okay to list a module under `packages`?

asamalik:
The [workload config spec](https://github.com/minimization/content-resolver/blob/master/config_specs/workload.yaml#L84) has a `modules_enable` field you can use. And then just list your packages in `packages` and they'll get pulled in from the module.

That's basically like doing

```
$ dnf module enable module:stream
$ dnf install package1 package2 ...
```

jiekang:
@asamalik should the module:stream be matching Fedora or RHEL 8 or is this something we can choose ourselves here?

omajid:
AFAIK, the input to ELN is Fedora. You should be using the name of the Fedora module stream here.

I don't know what the right thing to do is if you want the module name in ELN to be different that what it is in Fedora.

asamalik:
It takes content from Fedora, so both packages and modules are Fedora.

Module names are not propagated to the end result, just the package list, so having different module names won't be a problem here.

jiekang:
Thank you for the information. I've updated the configuration with the `modules_enable` field.

jiekang:
@asamalik Hi! After this PR it seems the content set for RHEL 9 still doesn't show anything for jmc.

Is this because I used the `arch_packages` section instead of `packages`?

asamalik:
@jiekang Hi!

Using `arch_packages` is totally fine if you only want those on a subset of architectures.

Looks like those packages don't exist in Fedora Rawhide.

```
# dnf module enable jmc:latest
... output omitted ...
# dnf install jmc
Last metadata expiration check: 0:00:30 ago on Tue Aug 11 15:49:11 2020.
All matches were filtered out by modular filtering for argument: jmc
Error: Unable to find a match: jmc
```

The tooling is supposed to report when this happens, but there is a bug with the reporting for arch-specific packages. I'll fix that ASAP. With the fix, you'll see the workload as failed with the following error message:

```
The following required packages are not available:
  - jmc
  - jmc-core
```

But in the meantime, can you please check the packages `jmc` and `jmc-core` in rawhide?

jiekang:
Ah I see, got it. Right, the rawhide builds are not successful at this time; that will be fixed shortly. Thank you for the explanation!

PR #35, opened 2020-06-25
We don't want those Python packages

hroncok:


hroncok:
Based on the content of the list, I've added "binary" RPM names, not component names. Is that correct?

configs/view-eln.yaml
```diff
@@ -75,6 +75,15 @@ data:
   # be pulled into a RHEL-like environment.
   - python3.6
 
+  # Miro Hrončok:
+  # The older Python version might accidentally be pulled in by not yet rebuilt packages
+  - python3.8
```
hroncok, line 80: As a side note, could you please rebuild glade-libs and unbound-libs with Python 3.9?

torsava:
+1

configs/view-eln.yaml
sgallagher: > As a side note, could you please rebuild glade-libs and unbound-libs with Python 3.9?

I will take care of this.

PR #37, opened 2020-07-01
Add sst_identity_management workloads

mkosek:


configs/sst_identity_management-idm-client-tools.yaml
```diff
@@ -0,0 +1,20 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: IdM Client Tools
+  description: General software for enrolling system with external Identity Management server
+  maintainer: sst_identity_management
+
+  packages:
+  - freeipa-healthcheck
```
fcami, line 9: I think freeipa-healthcheck would be better off in "FreeIPA Server - Domain Controller"

configs/sst_identity_management-idm-client-tools.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - sssd-tools
+  - realmd
+  - adcli
+  - krb5-workstation
+  - openldap-clients
+  - tlog
+  - gssproxy
+  - certmonger
+
```
fcami, line 18: I am not sure, but freeipa-client-epn should either be part of this list or in a standalone list tagged "FreeIPA applications". EPN can (and probably should) be deployed on a client rather than on a FreeIPA server, but is part of the deployment anyway.

mkosek: We can split FreeIPA applications later, right now I think it can be in sst_identity_management-idm-client-tools.yaml with other client tools. The main goal is to make sure it is included in ELN.

configs/sst_identity_management-idm-client-tools.yaml
mkosek: Ok, I added it there too.

PR #41, opened 2020-07-09
Split Front Door web console workload

martinpitt:
Create separate workloads for supported applications (podman and
composer), to better see and track their dependency trees.


martinpitt:
@dperpeet pointed out the evolution in https://tiny.distro.builders/workload--sst_front_door-web-console--fedora-empty-base--repository-fedora-eln--x86_64.html , and the dependency list is too large to track it sensibly. This should hopefully help.

PR #47, opened 2020-07-15
add sst_networking-core-network-management.yaml

thom311:
Adds NetworkManager and a wider range of packages, that are optional dependencies.

### DIFF
--- a/configs/sst_networking-core-network-management.yaml
+++ b/configs/sst_networking-core-network-management.yaml
@@ -1,0 +1,52 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Core Network Management
+  description: The Core Network Management tools for the local host. This includes NetworkManager and related tools.
+  maintainer: sst_networking
+  packages:
+  - ModemManager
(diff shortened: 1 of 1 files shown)


configs/sst_networking-core-network-management.yaml
```diff
@@ -0,0 +1,52 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Core Network Management
+  description: The Core Network Management tools for the local host. This includes NetworkManager and related tools.
+  maintainer: sst_networking
+  packages:
+  - ModemManager
```
tyll, line 8: Thanks. Sorting the list alphabetically might make it easier to lookup packages.

thom311: They are sorted, with `lang C` and vim's sort. Granted, that is case-sensitive. I wasn't aware that another locale would be preferable.

tyll: the locale is good, case-insensitive would be nice.

PR #52, opened 2020-07-20
Bulk add of CS Apps SST components

hhorak:
Data based on https://docs.google.com/spreadsheets/d/1gPHVLu8S4ITFxL3tslmsNktEMGYcrd2wHd4_E-e9Vj4/edit#gid=2026456282

### DIFF
--- a/configs/sst_cs_apps-apps-cronie.yaml
+++ b/configs/sst_cs_apps-apps-cronie.yaml
@@ -1,0 +1,13 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Cronie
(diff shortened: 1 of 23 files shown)


asamalik:
That's quite a list!

One technical note, the unwanted lists need to have:

```
```

at the start. It's a different file type. Here's [the spec](https://github.com/minimization/content-resolver/blob/master/config_specs/exclusion.yaml) for reference.

Can you please change those?

hhorak:
Sure, will fix.

asamalik:
Thanks!

PR #57, opened 2020-07-21
sst_networking-management-desktop: Fix iproute2/iproute name

marceloleitner:
It's actually packaged as 'iproute'.
@cathay4t @thom311


thom311:
Oh. You are right (of course).

lgtm!

PR #61, opened 2020-07-23
RDMA Stack workloads

Honggang-LI:
These workloads are based on RHEL-8 BaseOS and Appstream workloads. Please see inline comments in workloads for details.


configs/sst_nic_rdma-unwanted.yaml
```diff
@@ -0,0 +1,42 @@
+document: feedback-pipeline-workload
```
asamalik, line 1: This needs to be `document: feedback-pipeline-unwanted` for an exclusion list

configs/sst_nic_rdma-rhel-only.yaml
```diff
@@ -0,0 +1,43 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: RDMA Stack RHEL Only
+  description: RDMA Stack RHEL Only - Packages should be in ELN
+  maintainer: sst_nic_rdma
+
+  package_placeholders:
+    package-name: mvapich2
```
asamalik, line 9: You have a syntax error here. 

Instead of `package-name: mvapich2:`, it should only be `mvapich2:`

Like this:

```
package_placeholders:
  mvapich2:   # <-- that's how you set the package name correctly
    description: This package is not in Fedora (yet), but we want to see it here.
    requires:
    - environment-modules
    - hwloc-libs
 ...
```

configs/sst_nic_rdma-unwanted.yaml
Honggang-LI: Fixed. thanks

configs/sst_nic_rdma-rhel-only.yaml
Honggang-LI: Fixed too.

PR #73, opened 2020-07-25
sst_networking-management-server: add wpa_supplicant

marceloleitner:
It is already included in desktop workload, but as it is carrying
the names "server" and "desktop", lets make it explicit that this
package is also expected on server as well.

Noticed by Davide.

@thom311 @cathay4t


cathay4t:
Nice finding. Thanks.

PR #90, opened 2020-07-28
Added Platform Tools debugger packages.

drsmith2:
Adding Platform Tools debugger packages.


configs/sst_platform_tools-performance-tools-unwanted.yaml
```diff
@@ -0,0 +1,18 @@
+document: feedback-pipeline-workload
```
asamalik, line 1: Can you please change this to:

```
```

The exclusion list is a different document type.

drsmith2: I updated the document type.

asamalik: Thank you!

PR #127, opened 2020-08-03
Mark Tomcat as a removed package

cipherboy:
`Signed-off-by: Alexander Scheel <ascheel@redhat.com>`

---

/cc @csutherl -- Coty is the Fedora maintainer of Tomcat.


cipherboy:
Note that this helps with #126 -- Tomcat pulls in ECJ. In RHEL 8, an unsupported, limited-use partial replacement exists for Tomcat which does not pull in ECJ. That package is not in Fedora. It is likely that something similar will continue to exist in RHEL 9, where Tomcat will not be supported.

PR #130, opened 2020-08-03
Initial OpenSCAP-related workflows

dahaic:
SCAP and Compliance related workflows


dahaic:
FYI - I have made it a Draft PR, as I would like my colleagues to look at it first ;)

dahaic:
OK, seems to be correct for now.

PR #153, opened 2020-08-06
gcc-toolset: Add gcc-libraries to package_placeholders.

mpolacek:
We might need gcc-libraries to support applications built with the GCC Toolset, if a new toolset requires libraries not present in the system gcc.

### DIFF
--- a/configs/sst_platform_tools-gcc-toolset.yaml
+++ b/configs/sst_platform_tools-gcc-toolset.yaml
@@ -35,23 +35,24 @@ data:
   - chrpath
 
   package_placeholders:
-    gcc-libraries: This package might be needed to support applications built with the GCC Toolset.
-    requires:
-    - binutils
-    - glibc-devel
-    - libgcc
-    - libgomp
-    buildrequires:
-    - gcc
-    - gcc-c++
-    - gcc-gfortran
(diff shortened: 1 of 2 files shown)


configs/sst_platform_tools-gcc-toolset.yaml
```diff
@@ -34,5 +34,24 @@ data:
   - dwz
   - chrpath
 
+  package_placeholders:
+    gcc-libraries: This package might be needed to support applications built with the GCC Toolset.
+    requires:
```
asamalik, line 39: There needs to be one more indentation from here on.

The structure is:

```
  package_placeholders:
    package-name:
      description: This package is not in Fedora (yet), but we want to see it here.
      requires:
      - pkg1
      - pkg2
      buildrequires:
      - pkg2
      - pkg3
```

mpolacek: Oop, fixed in the following commit.

asamalik:
This has been opened for about a week now. I can't merge it before you make requested changes, please see my comment above.

asamalik:
Oh, sorry, you did that, I got confused, merging!

PR #176, opened 2020-08-10
2 more changes for the sst_kernel_rats team

jkacur:
rteval-loads was not showing up, I believe because the indentation in the yaml file was not consistent, so this fixes it.
rt-setup was not showing up, because it is not in fedora, and lacked a package-placeholders section, this fixes that.


asamalik:
Thanks!

I've already patched the tool to report non-existing packages.

PR #177, opened 2020-08-10
Add .NET 6.0

omajid:
Upstream expects to release this around Nov 2021. It will be an LTS release. It's probably a good idea to add it to ELN for now.

### DIFF
--- a/configs/sst_dotnet_java-dotnet.yaml
+++ b/configs/sst_dotnet_java-dotnet.yaml
@@ -5,13 +5,16 @@ data:
   description: .NET and .NET Core SDK, tools and runtime
   maintainer: sst_dotnet_java
   
-  packages: []
+  packages:
+  - dotnet-sdk-3.1
+  - dotnet-runtime-3.1
+  - aspnetcore-runtime-3.1
 
-  arch_packages:
-    x86_64:
-    - dotnet-sdk-3.1
-    - dotnet-runtime-3.1
-    - aspnetcore-runtime-3.1
+  package_placeholders:
+    dotnet-sdk-6.0:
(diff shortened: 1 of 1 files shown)


omajid:
@asamalik Can you help me understand the arch vs non-arch package distinction?

I want the future `dotnet-sdk-6.0` to be supported on the same architectures at the current `dotnet-sdk-3.1`. Is there an arch-specific `package_placeholders`?

dotnet-sdk-3.1's spec file says that it is `ExclusiveArch x86_64 aarch64`. If I move `dotnet-sdk-3.1` to (non-arch) `packages` in this commit, does that mean we want to include it for all Fedora/ELN architectures? Or just x86_64 and aarch64?

asamalik:
@omajid Sure!

Both `packages` and `package_placeholders` pull packages in for all available architectures.

There is also `arch_packages` (see [spec](https://github.com/minimization/content-resolver/blob/master/config_specs/workload.yaml#L60)) where you can define architecture-specific packages. So, like in your example, you could put `dotnet-sdk-3.1` in `arch_packages` under `x86_64` and  `aarch64` to only have it availabe on those two.

For `package_placeholders`, however, there's no such thing at the moment. Do you need that? I could add it.

omajid:
Thanks. Does that mean I can use `packages`/`package_placeholders` and ELN will inherit all/only the architectures that are used in Fedora? If the Fedora package is `x86_64` and `aarch64`, ELN wont try to pull that package in for, say, `s390x`, will it?

asamalik:
Exacly. It only pulls what's in Fedora's repos. If you try to include something that's not there, the workload fails and you get an error message.

omajid:
Thanks. I will leave all .NET packages as non-arch-specific then. AFAIK, the "supported" architectures are the same between ELN and Fedora.

PR #179, opened 2020-08-10
Add hesiod placeholder

frozencemetery:
Hesiod is not present in Fedora, and is marked deprecated in RHEL.  Accordingly, the binary package is called compat-hesiod.  It might be worth double-checking I got the source vs. binary distinction right in the yaml.


pnemade:
@asamalik This workload should have been created in `configs` directory right?

asamalik:
Yes, thanks for noticing, moved it!

Issue #162, opened 2020-08-07
python2.7, python3.6 and python3.8 still in ELN


hroncok:
Hey ELN people.

Despite my efforts to block python2.7, python3.6 and python3.8 from ELN, I still see that the packages are built there.

Could you please help me understand what needs to be done to prevent that from happening?

I've tried to communicate this in https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/U4TY26R3S65E474FUKGKB6H4JOCIEITY/ but there was no useful info except one private reply by @sgallagher who suggested that the proper communication channel for this are PRs in this repo.

The relevant PRs were:

- https://github.com/minimization/content-resolver-input/pull/25
- https://github.com/minimization/content-resolver-input/pull/35

Later moved around by https://github.com/minimization/content-resolver-input/pull/68.

tdawson:
Two answers.

First - marking something as unwanted does not automatically remove it.  It marks it as unwanted.
You can then go to the [content resolver](https://tiny.distro.builders/view--view-eln.html) and see the list of packages marked as unwanted that are still in the release.  There are also ways to track down what is pulling in those unwanted dependencies.
In a few weeks, (when the vast majority of the wanted packages are set) we will start working our way through those unwanted packages.  Opening pull requests and/or bugs.

Second - You have shown me I have a bug in my buildroot-generator script.
I re-wrote it specifically so that it wouldn't show duplicate packages (python3.5, 3.6, 3.7 and 3.8) for the same build dependencies.  But, I wasn't using python3.* as my test case, I was using fedora-release* ... and a couple others, as my test case.  And for those, this worked.
But, since all the different python's can be installed at the same time, they all get listed as build dependencies.
I'm going back to work on my script and get it fixed.

hroncok:
> First - marking something as unwanted does not automatically remove it. It marks it as unwanted...

Yes, I am aware of that. That's why I've said (in the email thread):

> While I understand that python2.7 might still be pulled in as dependency by
> other ELN packages (presumably gimp), I don't understand why is python3.8 and
> python3.6 built."

----

> I re-wrote it specifically so that it wouldn't show duplicate packages (python3.5, 3.6, 3.7 and 3.8)...

I am not sure I follow what "duplicate packages" means here.

> But, since all the different python's can be installed at the same time, they all get listed as build dependencies.

[...]

PR #180, opened 2020-08-10
Mark all alternate Python packages as unwanted

hroncok:
Related to https://github.com/minimization/content-resolver-input/issues/162

cc @torsava


torsava:
Good idea!

PR #185, opened 2020-08-11
Update sst_desktop(_applications) workflows

tpopela:
* Remove a totem subpackage that doesn't exist anymore
* Explicitly require exiv2 from GNOME Photos workflow
* Be explicit about all the required Qt 5 packages

### DIFF
--- a/configs/sst_desktop-totem.yaml
+++ b/configs/sst_desktop-totem.yaml
@@ -7,7 +7,6 @@ data:
 
   packages:
   - totem
-  - totem-nautilus
 
   labels:
   - eln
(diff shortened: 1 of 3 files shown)


tdawson:
The package 'qt5' has been removed from rawhide.
It is a metapackage that they removed in F32, but just happens to be built in ELN due to a bad %if in the spec file.

It would be a good idea to remove 'qt5' (just that one package) from configs/sst_desktop_applications-qt5.yaml

grulja:
The package 'qt5' has not been removed, it still exists to provide `qt5-rpm-macros` and `qt5-srpm-macros` subpackages. It just doesn't produce `qt5` and `qt5-devel` subpackages to install all modules and devel packages with a single package.

tpopela:
There was some confusion between me and @grulja. What @tdawson wrote still applies. I will create a PR to remove qt5 from there.

PR #193, opened 2020-08-12
sst_kernel_rats-setup.yaml: Add arch-packages section

jkacur:
- rt-setup has x86_64 as an exclusive arch, specify that in the yaml file
- add packages: [] in the main body since all the packages are listed in
the package placeholders
- Fix package_placeholders (underscore, not dash)
- Some reformatting and respacing

Signed-off-by: John Kacur <jkacur@redhat.com>


configs/sst_kernel_rats-setup.yaml
```diff
@@ -4,23 +4,25 @@ data:
...
   maintainer: sst_kernel_rats
-  packages:
-    - rt-setup
+  packages: []
   labels:
-    - eln
+  - eln
 
-  package-placeholders:
+  arch_packages:
```
asamalik, line 11: This probably won't do what you want. I assume you want to include the `rt-setup` as a placeholder only for x86_64, right? 

`arch_packages` pulls in packages from the repos the same way `packages` does. You need to add a `limit_arches` field under your placeholder, like this: https://github.com/minimization/content-resolver/blob/master/config_specs/workload.yaml#L102

jkacur:
Added a change to use limit_arches as requested

asamalik:
That looks good, thanks!

PR #200, opened 2020-08-18
rasdaemon: list every supported architecture

aristeu:
Since ExclusiveArch in the RPM isn't being used.

Signed-off-by: Aristeu Rozanski <arozansk@redhat.com>

### DIFF
--- a/configs/sst_kernel_ft-rasdaemon.yaml
+++ b/configs/sst_kernel_ft-rasdaemon.yaml
@@ -5,8 +5,7 @@ data:
   description: Utility to receive RAS error tracings
   maintainer: sst_kernel_ft
 
-  packages:
-  - rasdaemon
+  packages: []
 
   labels:      # <-- Please leave
   - eln        #     this unchanged.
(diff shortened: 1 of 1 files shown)


configs/sst_kernel_ft-rasdaemon.yaml
```diff
@@ -18,3 +18,16 @@ data:
   #           - arch-specific-package1   # <-- The list of binary RPM packages
   #               - arch-specific-package1   #     goes here.
   #
+  arch_packages:
+      x86_64:
+          - rasdaemon
```
asamalik, line 23: I see you're listing `rasdaemon` as an arch-specific package, but it's also in the `packages` section which would pull in into all arches again.

You might might want to use:

```
packages: []
```

aristeu: That was my guess to fix:
https://tiny.distro.builders/workload-overview--sst_kernel_ft-rasdaemon--repository-fedora-eln.html

Apparently ExclusiveArch is being ignored and it's being tested in s390.

PR #202, opened 2020-08-18
mark kpatch package as RHEL-only

joe-lawrence:
Signed-off-by: Joe Lawrence <joe.lawrence@redhat.com>

### DIFF
--- a/configs/sst_kernel_ft-kpatch.yaml
+++ b/configs/sst_kernel_ft-kpatch.yaml
@@ -5,14 +5,14 @@ data:
   description: Dynamic kernel patch manager
   maintainer: sst_kernel_ft
 
-  packages:
-  - kpatch
+  packages: []
 
   labels:      # <-- Please leave
   - eln        #     this unchanged.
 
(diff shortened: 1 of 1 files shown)


configs/sst_kernel_ft-kpatch.yaml
```diff
@@ -11,8 +11,9 @@ data:
...
   - eln        #     this unchanged.
 
-#  # (Only include the following section if you have arch-specific packages.)
-#  arch_packages:
-#    x86_64:                    # <-- Architecture
-#    - arch-specific-package1   # <-- The list of binary RPM packages
-#    - arch-specific-package1   #     goes here.
+  package_placeholders:
+    # kpatch is a RHEL-only package
+    kpatch:
```
asamalik, line 16: You'll also need to remove it from the `packages` field.

You can do

```
packages: []
```

joe-lawrence: v2: specify empty packages set as @asamalik suggested

PR #207, opened 2020-08-18
SST_FILESYSTEMS: Adjusted content set definition for rpcsvc-proto.

rhawalsh:
The rpcsvc-proto component does not provide a package with the basename of the
spec file (rpcsvc-proto).  This can be confirmed by looking at one of the
recent koji builds [0].

[0] https://koji.fedoraproject.org/koji/buildinfo?buildID=1577169


rhawalsh:
This is obviated by [pull 206](https://github.com/minimization/content-resolver-input/pull/207).

PR #208, opened 2020-08-19
sst_kernel_maintainers: add package_placeholders for packages Fedora doesn't build

jstancek:
Signed-off-by: Jan Stancek <jstancek@redhat.com>


jstancek:
https://tiny.distro.builders/maintainer--sst_kernel_maintainers.html shows success, so this appear to be no longer needed.

PR #209, opened 2020-08-19
sst_networking: add package placeholder for "NetworkManager-config-co…

thom311:
…nnectivity-redhat"

The "NetworkManager-config-connectivity-redhat" package is a subpackage
of the "NetworkManager" source RPM.

On RHEL we build this package, but on Fedora there is instead
"NetworkManager-config-connectivity-fedora". This causes the workload to
be invalid.

Fix that by adding it as "package_placeholders".

Note that since the package is a subpackage, it has essentially the same
build requirements as "NetworkManager".

### DIFF
--- a/configs/sst_networking-management-server.yaml
+++ b/configs/sst_networking-management-server.yaml
@@ -33,7 +33,6 @@ data:
   - NetworkManager-adsl
 
   # Config files
-  - NetworkManager-config-connectivity-redhat
   - NetworkManager-config-server
 
   # Cellular Modem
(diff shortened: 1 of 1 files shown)


tyll:
Is this the correct solution? I understood that there are RPM macros in ELN that allow to build different than in Fedora, so maybe the NM spec needs to be updated to take this into account.

tyll:
Actually the macros seem to be correct, since it is properly reported in https://tiny.distro.builders/workload--sst_networking-management-server--fedora-empty-base--repository-fedora-eln--x86_64.html

The error occurs when it is tested in Rawhide:
https://tiny.distro.builders/workload--sst_networking-management-server--fedora-empty-base--repository-fedora-rawhide--x86_64.html

Since the workload is tagged as ELN, why is it tested against Rawhide?

asamalik:
@tyll Fair, I think we could remove the rawhide results now. They were useful when we were bootstrapping ELN. It gave us something rough to start with.

asamalik:
@thom311 So, based on what @tyll has pointed out, and looking at the results, the ELN results are fine, the `NetworkManager-config-connectivity-redhat` package is present there. Do you still want to have this merged anyway?

thom311:
ok, I think @tyll is right. Thanks.

I will close this pull-request for now. Maybe there is nothing to do, or we'll need to find a better solution.

marceloleitner:
Hah, and I had thought/assumed NetworkManager-config-connectivity-redhat is package of its own.
Thanks.

PR #217, opened 2020-08-20
sst_kerenl_rats-perf: Limit rt-tests to x86_64 architecture

jkacur:
Limit the rt-tests suite to the x86_64 architecture.

Signed-off-by: John Kacur <jkacur@redhat.com>

### DIFF
--- a/configs/sst_kernel_rats-perf.yaml
+++ b/configs/sst_kernel_rats-perf.yaml
@@ -20,6 +20,8 @@ data:
       - make
       - numactl-devel
       - python3-devel
+      limit_arches:
+      - x86_64
     rteval:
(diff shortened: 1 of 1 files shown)


configs/sst_kernel_rats-perf.yaml
```diff
@@ -20,6 +20,8 @@ data:
       - make
       - numactl-devel
       - python3-devel
+    limit_arches:
```
asamalik, line 23: That needs one more level of indentation, so it's under `rt-tests`. (Would do that for you, but can't edit your branch)

asamalik (changes requested):

jkacur:
Oops, indentation fixed, thanks for the catch.

PR #225, opened 2020-08-24
Add pki-servlet-engine as a placeholder package

cipherboy:
This package is only available in RHEL8+ and is present in Fedora under
the tomcat package name. Note that `tomcat` package is unwanted in RHEL8+.

See-Also: "tomcat in RHEL 9" with Adam Samalik, Tomas Tomecek, Coty
Sutherland, Troy Dawson, and myself.

`Signed-off-by: Alexander Scheel <ascheel@redhat.com>`

---

@asamalik -- do I need to add this to _each_ IPA workload that requires Tomcat? Including `sst_identity_management-freeipa-server-full.yaml` and `sst_identity_management-freeipa-server-dc.yaml`? Or is once enough? 

/cc @csutherl @mkosek

### DIFF
--- a/configs/sst_identity_management-freeipa-server.yaml
+++ b/configs/sst_identity_management-freeipa-server.yaml
@@ -10,3 +10,24 @@ data:
 
   labels:
   - eln
+
+  package_placeholders:
+    pki-servlet-engine:
+      description: This package provides a Tomcat-like servlet engine in RHEL 8+
+      requires:
+       - ant
+       - java-headless
+       - java-devel
+       - jpackage-utils
+       - procps
+       - shadow-utils
+       - chkconfig
+       - systemd-units
+       - coreutils
+      buildrequires:
+       - ant
+       - findutils
+       - java-devel
+       - jpackage-utils
+       - maven-local
+       - systemd-unnits


mkosek:
I assume that this package is not needed everywhere, to avoid repetition and thus risk of inconsistency. But I would let @asamalik to tell us if this approach is OK.

asamalik:
This will probably not do what you intend to do.

`package_placeholders` just list an arbitrary package name on the output, and pull in dependencies they have listed. But they won't be part of the dependency resolution themselves.

`package_placeholders` are useful to express, for example, a package that's coming to Fedora at some point, but it's not there yet, but we want to make sure the dependencies are there already.

So this will add a `pki-servlet-engine` package as required to the list on the output, and its listed dependencies will be pulled in for the DNF transaction. But because `pki-servlet-engine` won't be in the transaction (it doesn't exist), other packages that need `tomcat` will still pull it in.

asamalik:
An approach that would work would be to add packages pulling `tomcat` as `package_placeholders`. Otherwise we'll need to figure out a different approach.

cipherboy:
@asamalik I was going to add this in addition to this Tomcat PR: https://src.fedoraproject.org/rpms/tomcat/pull-request/6 

My assumption below is that Dogtag are the only packages depending on `tomcat`, thus pulling it into RHEL 9.

This means `tomcat` SRPM will advertise itself as `pki-servlet-engine` in ELN. This will let `tomcatjss` and `dogtag-pki`/`pki-core` rebuild (and thus switch from `Requires: Tomcat` to `Requires: pki-servlet-engine` in ELN). Then, if the combination of `package_placeholders` + `unwanted_packages` (in #127) works correctly, `tomcat` SRPM won't get pulled in but some mythical `pki-servlet-engine` from `package_placeholders` will.

This also has the nice side effect of not requiring introducing a "real" `pki-servlet-engine` in Fedora/Fedora ELN, while still letting Fedora ELN work fine (because `tomcat` will provide the real SRPM).

It might require bug fixes in `content-resolver` to actually behave like described above, but I think that makes sense.

cipherboy:
Closing in favor of modularizing Fedora ELN per discussion with @sgallagher.

PR #246, opened 2020-08-28
sst_kernel_rats - limit packages in kernel-rt and perf to x86_64

clrkwllms:


configs/sst_kernel_rats-kernel-rt-base.yaml
```diff
@@ -12,6 +12,8 @@ data:
         - tuned-profiles-nfv-host-bin
     labels:
         - eln
+    limit_arches:
```
asamalik, line 15: This won't work. If you want to make all the package s above to only be on x86_64, you can move them under:

```
 arch_packages:
     x86_64:
       - tuned-profiles-realtime
       ...
```

and leave the packages field empty:

```
packages: []
```

https://github.com/minimization/content-resolver/blob/master/config_specs/workload.yaml#L61

asamalik (changes requested):

configs/sst_kernel_rats-kernel-rt-base.yaml
```diff
@@ -5,11 +5,6 @@ data:
     description: Base set of kernel-rt packages and tools to ship
     maintainer: sst_kernel_rats
     packages:
```
asamalik, line 7: This field is missing value. You can do `packages: []` to leave it empty.

configs/sst_kernel_rats-kernel-rt-base.yaml
```diff
@@ -243,3 +238,44 @@ data:
...
         limit_arches:
           - x86_64
+
+      tuned-profiles-realtime:
+        description: This package is not in Fedora (yet), but we want it here.
+        requires:
+          - bash
+          - man
+        limit_arches:
+          - noarch
```
asamalik, line 248: `noarch` is not a valid architecture here. This field is not the "arch" part of the RPM name, it determines for what actual architectures is this package present.

asamalik (changes requested):

configs/sst_kernel_rats-kernel-rt-base.yaml
clrkwllms: Ok, I'll drop all those noarch sections

configs/sst_kernel_rats-kernel-rt-base.yaml
clrkwllms: Didn't realize that I needed the empty brackets.   I Hate YAML (tm)

configs/sst_kernel_rats-perf.yaml
```diff
@@ -12,6 +12,8 @@ data:
   - rteval-loads
   labels:
   - eln
+  limit_arches:
+    - x86_64
```
asamalik, line 16: This won't work. If you want to make all the package s above to only be on x86_64, you can move them under:

```
 arch_packages:
     x86_64:
       - tuned-profiles-realtime
       ...
```

and leave the packages field empty:

```
packages: []
```

The `limit_arches` is only available for `package_placeholders`.

See the [workload spec](https://github.com/minimization/content-resolver/blob/master/config_specs/workload.yaml#L61) for all available fields.

asamalik (changes requested):
Hey, sorry, I don't want to torture you with constant requests for changes. I believe you simply want all of these workloads to be `x86_64` only? 

If so, all existing packages need to be listed under:

```
arch_packages:
    x86_64:
      - package-name
```

... and all non-existing packages under `package_placeholders` need to have the...

```
limit_arches:
    - x86_64
```

... property. Many of those already have that, but not all.

clrkwllms:
Thanks, I'll wait for the next run here:
https://github.com/minimization/content-resolver-input and work through the
errors.

PR #247, opened 2020-09-01
Replace glibc-headers-x86_64 with glibc-headers-x86

fweimer:
Fixes commit eeb21a2ef008ba5353d14ef4c717864ca8fe1ce9 ("Add Platform
Tools C and C++ Runtime/Devel.") and commit
107af77ef635c4975283e09a41e5614b50356fe3 ("Add Platform Tools C
Runtime and Development.").

Suggested by Troy Dawson.

### DIFF
--- a/configs/sst_platform_tools-c++-development-full.yaml
+++ b/configs/sst_platform_tools-c++-development-full.yaml
@@ -18,7 +18,7 @@ data:
   - libgcc
   # Noarch header packages.
   - glibc-headers-s390
-  - glibc-headers-x86_64
+  - glibc-headers-x86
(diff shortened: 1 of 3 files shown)


fweimer:
The larger question is whether we actually have to list any internal packages here, or can rely on the dependency analyzer to pull them in.

asamalik:
@fweimer The tool pulls all dependencies in, so this is up to you. Listing packages explicitly vs. letting them being pulled in helps us determine the reason of why a package is present — do we actually need it, or is it just a dependency? And then we can make other decisions based on that when minimising the set.

PR #249, opened 2020-09-01
Add initial content set definition for Virtualization SST

yashkmankad:
Signed-off-by: Yash Mankad <ymankad@redhat.com>


configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - ipxe
+  - kernel
+  - libdnet
+  - libguestfs
+  - libguestfs-winsupport
+  - libiscsi
+  - libmspack
+  - libnbd
+  - libosinfo
+  - libunwind
```
jnmnus, line 33: libunwind should be removed. It was dropped in RHEL-8.

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  modules_enable:
+  - virt:rhel
+
+  packages:
+  - amtterm
+  - augeas
+  - dtc
+  - edk2
+  - fakechroot
+  - fakeroot
```
rwmjones, line 17: Arrrggghh `fakechroot` and `fakeroot` are the packages which never die.  Please remove these two.

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - hivex
+  - icoutils
+  - ipxe
+  - kernel
+  - libdnet
+  - libguestfs
+  - libguestfs-winsupport
+  - libiscsi
+  - libmspack
+  - libnbd
```
rwmjones, line 31: I would really like `libnbd` to appear in AppStream in RHEL 9.  (In RHEL 8 it is in the Virt module, but that was a mistake in hindsight).

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - mingw-gcc
+  - mingw-headers
+  - mingw-pkg-config
+  - mingw-srvany
+  - mingw-w64-tools
+  - nbdkit
+  - netcf
+  - ocaml
+  - ocaml-calendar
+  - ocaml-camlp4
```
rwmjones, line 52: `ocaml-camlp4` is dead upstream and should not appear in RHEL 9.

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - qemu-guest-agent
+  - qemu-kvm
+  - rhevm-guest-agent
+  - seabios
+  - sgabios
+  - SLOF
+  - spice-html5
+  - spice-xpi
+  - supermin
+  - vhostmd
```
rwmjones, line 81: In RHEL 8 `vhostmd` is in BaseOS I think.  It's used for SAP customers only.

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - virt-manager
+  - virt-p2v
```
rwmjones, line 83: `virt-p2v` is shipped as a separate ISO.  We don't normally list it as a separate package (eg. it doesn't appear in the RHEL 8 virt module).

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - virt:rhel
+
+  packages:
+  - amtterm
+  - augeas
+  - dtc
+  - edk2
+  - fakechroot
+  - fakeroot
+  - fuse-sshfs
```
rwmjones, line 18: I've long wanted `fuse-sshfs` to be in RHEL, championing it since RHEL 5, but we never managed to get it in.  It's something that lots of customers want though.  However it's not a virt module package, probably AppStream.

rwmjones:
`python-pyvmomi` appears in the RHEL 8 virt module but not in this list.

rwmjones:
`open-vm-tools` is not on this list, but its dependencies (`libdnet`, `libmspack`) are.

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  name: Virtualization on RHEL - KVM Userspace
+  description: Userspace agents and tools that enable virtualization capabilities on RHEL
+  maintainer: sst_virtualization
+
+  modules_enable:
+  - virt:rhel
+
+  packages:
+  - amtterm
+  - augeas
```
ptoscano, line 13: augeas is part of the SST because I'm its maintainer. augeas is in BaseOS, outside of any module.

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - dtc
+  - edk2
+  - fakechroot
+  - fakeroot
+  - fuse-sshfs
+  - gperftools
+  - gtk-vnc
```
berrange, line 20: This is needed by desktop apps, so can't be in the virt module AFAIK

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - libvirt-snmp
+  - mdevctl
+  - mingw-binutils
+  - mingw-crt
+  - mingw-filesystem
+  - mingw-gcc
+  - mingw-headers
+  - mingw-pkg-config
+  - mingw-srvany
+  - mingw-w64-tools
```
berrange, line 47: None of the mingw stuff should be listed under virt

configs/sst_virtualization-virt.yaml
yashkmankad: If it will still be maintained by you (or someone from sst_virtualization), then I would leave it here, as I don't think file differentiates between modular and non-modular content.

@asamalik - is there way for me to specify libnbd in the way Rich is asking ? Keep it in sst_virt, but not in the virt module.

configs/sst_virtualization-virt.yaml
yashkmankad: But should they be in RHEL-9 ?

If yes, we need to ask the sst_desktop (or the right SST) to take over ownership and mention these in their Content Set. 
If no, they will go in the  exclusion list.

rwmjones: Note that `mingw-srvany` is not like the other packages.  It is a binary that is used by libguestfs and virt-v2v.

configs/sst_virtualization-virt.yaml
```diff
@@ -0,0 +1,90 @@
...
+  - hexedit
+  - hivex
+  - icoutils
+  - ipxe
+  - kernel
+  - libdnet
+  - libguestfs
+  - libguestfs-winsupport
+  - libiscsi
+  - libmspack
```
yashkmankad, line 28: @rwmjones should libmspack and libdnet be moved to https://github.com/minimization/content-resolver-input/blob/master/configs/sst_virtualization_cloud-3rdparty_hypervisors.yaml ?

If yes, I ll ask Vitaly to update the file.

rwmjones: Probably best to ask Cathy any questions related to `open-vm-tools` and its dependencies.  Note that `xmlsec1` is another depdendency of this package.

configs/sst_virtualization-virt.yaml
rwmjones: So in addition to what I said above, please don't remove `mingw-srvany`.  You can assign it to me if you like.

asamalik:
BTW if you want a preview of what you have now, I can merge it and you can then keep tweaking it. It doesn't need to be final at the first shot. That way you'll see what dependencies are pulled in, changes might become more obvious, etc.

yashkmankad:
@asamalik I ll just be pushing a couple of changes in the next 30 mins - 1 hour. I will let you know once I have done so, then we should be good to merge the pull request. Thanks!

configs/sst_virtualization-virt.yaml
yashkmankad: The BaseOS / AppStream split will be happening later. 
So this will stay here for now.

configs/sst_virtualization-virt.yaml
yashkmankad: Is it a rpm though ? Or only an ISO ?
This is just to determine whether we want a package shipped in RHEL-9, and not specifically the virt module.

configs/sst_virtualization-virt.yaml
yashkmankad: The BaseOS / Appstream split happens later. So as long as this is maintained by SST Virt, it stays here for now.
Same for `vhostmd` and `fuse-sshfs`.

`python-pyvmomi` is a dependency and will be pulled in by the content resolver tool. I will keep an eye for it after the initial resolver tool runs for Virt.

configs/sst_virtualization-virt.yaml
yashkmankad: We are trying to determine the list of packages that will be owned by the Virt SST, and not the virt module. 
Does `gtk-vnc` below with sst_virt ?or should it be moved to another SST ?

yashkmankad:
@asamalik the PR can be merged now.
We will make further changes with additional PRs as required.

configs/sst_virtualization-virt.yaml
rwmjones: It's both in fact.  We ship an ISO, but also an RPM containing the tools to make the ISO, both called `virt-p2v`.

PR #250, opened 2020-09-01
[Do Not Merge] Add initial content set definition for Advanced Virt SST

yashkmankad:
Signed-off-by: Yash Mankad <ymankad@redhat.com>

### DIFF
--- a/configs/sst_virtualization_advanced-virt.yaml
+++ b/configs/sst_virtualization_advanced-virt.yaml
@@ -1,0 +1,46 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Advanced Virtualization on RHEL - for Red Hat Layered Products
+  description: Userspace agents and tools that enable advanced virtualization capabilities on RHEL
+  maintainer: sst_virtualization_advanced
+
+  modules_enable:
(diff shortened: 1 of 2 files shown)


configs/sst_virtualization_advanced-virt.yaml
```diff
@@ -0,0 +1,46 @@
...
+  packages:
+  - hivex
+  - kata-agent
+  - kata-osbuilder
+  - kata-runtime
+  - kata-shim
+  - libguestfs
+  - libguestfs-winsupport
+  - libiscsi
+  - libnbd
```
rwmjones, line 20: I would really like `libnbd` to appear in AppStream in RHEL 9.  It is in RHEL 8 virt module but that was a mistake.

configs/sst_virtualization_advanced-virt.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - libssh2
```
rwmjones, line 21: No no no ... `libssh2` must not appear in RHEL 9.  `libssh` replaces it.

yashkmankad:
Discussion deferred.

PR #260, opened 2020-09-04
address resolver errors for perf and devel/debug workloads 

clrkwllms:
Fix a package name typo
Try moving the kernel-rt-debuginfo-common-x86_64 block in an attempt to address resolver failure to see it

### DIFF
--- a/configs/sst_kernel_rats-kernel-rt-develdebug.yaml
+++ b/configs/sst_kernel_rats-kernel-rt-develdebug.yaml
@@ -239,6 +239,18 @@ data:
         limit_arches:
           - x86_64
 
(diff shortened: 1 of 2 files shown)


asamalik:
@clrkwllms Right now placeholders can't depend on another placeholders. But that's not an issue, because since you're listing them, they'll be present anyway. So that's why I did my fix by removing the dependency record. Otherwise the list isn't ordered, so moving them won't have any effect.

But let's fix the typo, merging!

clrkwllms:
Honestly, moving the block was a "shouldnt' work but might as well try it",
thank you for pulling the dependency.

That means the next run that happens, we should see no errors!

PR #264, opened 2020-09-07
Do not explicitly list glibc-headers-x86, glibc-headers-s390

fweimer:
These packages are available on all architectures, but are only
installable on x86_64 and s390x.  On other architectures, they
have file conflicts with glibc-devel.

Red Hat GmbH, https://de.redhat.com/ , Registered seat: Grasbrunn,
Commercial register: Amtsgericht Muenchen, HRB 153243,
Managing Directors: Charles Cachera, Brian Klemm, Laurie Krebs, Michael O'Neill

### DIFF
--- a/configs/sst_platform_tools-c++-development-full.yaml
+++ b/configs/sst_platform_tools-c++-development-full.yaml
@@ -16,9 +16,6 @@ data:
   - libstdc++
   - libstdc++-devel
   - libgcc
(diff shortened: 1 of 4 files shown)


asamalik:
If you want to include each on just one architecture, you can use the "arch_packages" field. Otherwise, looks, good, merging!

PR #265, opened 2020-09-09
Add initial split

jwboyer:
List some of the intial content split packages for the various
repos.  We'll start small and adjust from here.

### DIFF
--- a/configs/eln-repo-split.yaml
+++ b/configs/eln-repo-split.yaml
@@ -1,0 +1,168 @@
+document: eln-repo-split
+version: 1
+data:
+
+  name: Initial content split
+  maintainer: jwboyer
+
+  baseos:
+    must:
+    - kernel
+    - grub2
+    - glibc
+    - bash
+    - rpm
+    - dnf
+    - kmod
+    - linux-firmware
+    - curl
+    - openssh-clients
+    - openssh-server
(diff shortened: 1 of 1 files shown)


configs/eln-repo-split.yaml
```diff
@@ -0,0 +1,168 @@
...
+    - curl
+    - openssh-clients
+    - openssh-server
+    - policycoreutils
+    - selinux-policy
+    - selinux-policy-targeted
+    - systemd
+
+  appstream:
+    must:
```
asamalik, line 27: @jwboyer are you ok with changing this to `want`? The appstream list doesn't have a `must` as it's the default list, and it could result in conflicting or breaking results if something pulled to baseos would require something that would need to be suddenly forced in appstream.

asamalik (changes requested):

configs/eln-repo-split.yaml
jwboyer: Ah.  Yes.  Do you want me to change and resubmit?

jwboyer: I rebased and fixed this.  I have no idea if that's what I was supposed to do, but I did! :smile:

asamalik (approved):
Looks good, thanks!

PR #274, opened 2020-09-18
Add missing packages to SELinux workloads

bachradsusi:
And also rename the maintainer to sst_security_selinux

Please let me know when this change is propagated so I can re-check the diff between rhel8 and this.


asamalik:
It usually takes about 2 hours for the page to refresh

PR #275, opened 2020-09-18
Add: atlas s390x subpackages

imartisko:


configs/sst_cs_plumbers-math.yaml
```diff
@@ -0,0 +1,18 @@
...
+version: 1
+data:
+  name: Math:
+  description: Various math related libraries
+  maintainer: sst_cs_plumbers
+
+  packages:
+  - atlas
+  - atlas-devel
+    s390x:
```
asamalik, line 11: You're missing the `arch_packages` field above `s390x`.

asamalik (changes requested):

configs/sst_cs_plumbers-math.yaml
imartisko: Thanks, should be fixed now

PR #286, opened 2020-09-21
sst_rhgs-unwanted: remove glusterfs RPMs from ELN

kalebskeithley:
Resolution of where glusterfs libs and -devel packages come from
in RHEL 9 is required.

Josh Boyer and I have been having a discusion on the hallway track.
This is intended to surface the issue.


jwboyer:
Looks good to me.  Thanks!

PR #344, opened 2020-10-15
We need to start watching transitive hull of dependencies of some groups.

adelton:


asamalik:
Can you please add:

```
labels:
- fedora
```

... to all the configs? That currently matches them with an empty rawhide environment in which they'll be monitored.

adelton:
I thought leaving `labels` out will watch in all environments available. We'd like rawhide and ELN, at least -- would that be `fedora` and `eln` labels?

asamalik:
Empty labels would make it do nothing actually.

For ELN, we can't add the "eln" label because that would also add it to the set. But I'll set up a new one for you. Can you add a "eln-analyze-only" label? I'll do the rest.

adelton:
Ah, good. I've added the two labels in f7927204bc0f43460287cc7955a2df92c6036ebe now.

asamalik:
Looks good!

PR #348, opened 2020-10-17
Mark the Jupyter Notebook stack as unwanted

hroncok:


asamalik:
@hroncok BTW, you might find this useful, I recently added support for unwanted SRPMs: https://github.com/minimization/content-resolver/blob/master/config_specs/exclusion.yaml#L60

PR #353, opened 2020-10-21
Add more Software Mgmt pkgs

dmnks:


hroncok:
Hey @dmnks, could you please share more info why is kernel-rpm-macros needed when it is not BuildRequired? E.g. in https://src.fedoraproject.org/rpms/redhat-rpm-config/pull-request/128#comment-72506 -- thanks.

dmnks:
It was actually split to make the maintenance easier for the kernel folks, AIUI. Let me ask them first if they'd be interested in taking it under their wings.

And thanks for bringing this up - I originally assumed it would remain ours (like `kernel-srpm-macros`, though that one is a subpackage of `redhat-rpm-config`, not a separate SRPM component), but checking with Panu, it seems it really should be the kernel team instead.

dmnks:
Good question - I assumed it was needed to build kernel modules and since it was also available in RHEL-8, I didn't consider it something to be removed (for no good reason).

That said, let me follow up with the kernel folks to see what they think. I'm all for removal if it's of no use.

PR #357, opened 2020-10-22
coreos: Add bootupd

cgwalters:
See https://github.com/coreos/fedora-coreos-config/commit/8be272008a8ca58cc0ec53fb8301b50df3b1d899#diff-5d2a56392c73a744578b94f48544e03031836077d7030313d655f6d688687c94

### DIFF
--- a/configs/sst_coreos.yaml
+++ b/configs/sst_coreos.yaml
@@ -17,6 +17,7 @@ data:
    - runc
    - ssh-key-dir
    - zincati
+   - rust-bootupd
 
    # These packages are inherited from the base
    - cloud-utils-growpart


cgwalters:
I'm not totally sure if this is right (is the list binary or source packages)?

Is this all that's needed to get this built in ELN?

asamalik:
It's a binary list.

And yes, this is all to get it into ELN and built for you.

I'll let you confirm before merging (because you asked if it's binary or source, want to make sure it does what you expect).

cgwalters:
Yep, it *looked* like binaries to me so that's what I added.  Should be good to merge then, thanks!

darkmuggle:
/lgtm
/approve
This looks correct to me.

cgwalters:
Did this break the Content Resolver? 
https://tiny.distro.builders/workload--sst_coreos--fedora-empty-base--repository-fedora-eln--x86_64.html

Is there some sort of deadlock where the package has to built before the resolver can find it, and the ELN "build jenkins job" is waiting for the resolver to provide it?

(What's the linkage between those two?)

asamalik:
It uses Rawhide as a backup. So this means that package is not in Rawhide (or ELN of course).

cgwalters:
> It uses Rawhide as a backup. So this means that package is not in Rawhide (or ELN of course).

But https://koji.fedoraproject.org/koji/buildinfo?buildID=1634529

asamalik:
There's no `rust-bootupd` RPM on that page. Content Resolver needs RPMs, not SRPMs,

cgwalters:
Oh wow gah, we had that whole conversation and I *still* managed to put the SRPM in there.  I think I had it locally but forgot to amend.

cgwalters:
https://github.com/minimization/content-resolver-input/pull/362

asamalik:
No worries :D

PR #362, opened 2020-10-27
coreos: Fix bootupd name

cgwalters:
And while we have the patient open, clarify the comment.

### DIFF
--- a/configs/sst_coreos.yaml
+++ b/configs/sst_coreos.yaml
@@ -8,8 +8,9 @@ data:
   maintainer: sst_coreos
 
   packages:
-   # These packages are built and releases as part of
-   # Openshift.
+   # These packages are currently built and released as part of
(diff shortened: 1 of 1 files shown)


cgwalters:
How long until https://tiny.distro.builders/workload--sst_coreos--fedora-empty-base--repository-fedora-eln--x86_64.html is fixed?

asamalik:
I don't see the error there anymore, so, now. :)

asamalik:
Otherwise it refreshes every ~80 mins.

cgwalters:
Ah OK and looking at that page I see this broke ppc64le/s390x; is there an equivalent of [rpm-ostree architecture conditionals](https://github.com/coreos/fedora-coreos-config/blob/eef052a9e03959f012585dc89e6575fb00f2dfd9/manifests/fedora-coreos-base.yaml#L172) here?

PR #369, opened 2020-11-09
Explicitly mark libcgroup as unwanted

nforro:
libcgroup doesn't support cgroupv2 and has been effectively dead for a long time.

However, I can see it being listed as [openstack dependency](https://github.com/minimization/content-resolver-input/blob/01838f97eb7c542ac500349b3a0eca00f229d70d/configs/sst_openstack.yaml#L47). @jwboyer Do you happen to know why is it there/where it comes from?


jwboyer:
It comes from how they consume various packages today.  We can adjust going forward.

PR #370, opened 2020-11-11
accel-config: Add accel-config and subpackages to ELN

snits:
Signed-off-by: Jerry Snitselaar <jsnitsel@redhat.com>

### DIFF
--- a/configs/sst_kernel_rats-dma-iommu.yaml
+++ b/configs/sst_kernel_rats-dma-iommu.yaml
@@ -1,0 +1,15 @@
+document: feedback-pipeline-workload
+version: 1
+data:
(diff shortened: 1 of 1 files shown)


jwboyer:
This is only available on x86_64, correct?  If so, you probably need to add it via an archful package listing.  Something like:

arch_packages:
  x86_64:
    - accel-config
    - accel-config-devel
    - accel-config-libs

Otherwise the workload will get errors for the architectures that this isn't built for.

snits:
Oops. Yes, it should be be x86 only. Let me fix that up.

asamalik:
Looks good!

PR #372, opened 2020-11-12
We don't want this

hroncok:
cc @torsava


torsava:
@hroncok Love the commit message :) +1 Thanks!

asamalik:
It's super clear to me! :D

PR #373, opened 2020-11-13
Add git-lfs placeholder

opohorel:
Git-lfs pulls a lot of golang dependencies that we want to bundle in RHEL 9.
Making a conditional in Fedora package would make spec file larger and worse readable.
We decided to make git-lfs a "package placeholder" and pull only dependencies that we want.


voxik:
@hhorak FYI

PR #386, opened 2020-11-25
Use rubygem-tilt placeholder to reduce Ruby dependency chain.

voxik:
Tilt has gazillion dependencies and what is worser, half of them are circular. Luckily, they test just possible runtime dependencies which will be unused at the end. They are enabled on Fedora, just because they can be enabled. Therefore cut them off to reduce the Ruby dependency chain.

Also note that when dependency is not available, the tests are just skipped.

### DIFF
--- a/configs/sst_cs_apps-ruby.yaml
+++ b/configs/sst_cs_apps-ruby.yaml
@@ -11,3 +11,13 @@ data:
 
   labels:
   - eln
+
+  package_placeholders:
+    rubygem-tilt:
+      description: Generic interface to multiple Ruby template engines
+      requires:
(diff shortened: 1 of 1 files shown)


asamalik:
I don't think it did what you wanted.

The original `rubygem-tilt ` still gets pulled in: https://tiny.distro.builders/view-rpm--view-eln--rubygem-tilt.html

voxik:
Interesting. Is the placeholder name added among provides or do I need to provide it explicitly?

voxik:
Actually, there is not `provides: rubygem(tilt)`, that might be the issue.

PR #388, opened 2020-11-26
Add a placeholder for pip to filter out many tests deps

hroncok:
cc @torsava @encukou 

@asamalik Am I doing it right? python-pip is the source name, the other two are subpackages.

### DIFF
--- a/configs/sst_cs_apps-python.yaml
+++ b/configs/sst_cs_apps-python.yaml
@@ -14,3 +14,31 @@ data:
 
   labels:
   - eln
+
+  package_placeholders:
+    python3-pip:
+      description: We will build this without tests in RHEL 9
+      buildrequires:
+        - bash-completion
+        - ca-certificates
+        - python3-devel
+        - python3-setuptools
+        - python3-rpm-generators
+        - python3-wheel
+      requires:
+        - /usr/bin/python3
+        - ca-certificates
+        - python(abi) = 3.9
+      srpm: python-pip
+    python-pip-wheel:
+      description: We will build this without tests in RHEL 9
+      buildrequires:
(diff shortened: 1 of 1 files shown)


hroncok:
I've amended this based on https://github.com/minimization/content-resolver/blob/master/config_specs/workload.yaml

No idea if this is how to have 2 subpackages.

asamalik:
@hroncok yes, you're doing it right

hroncok:
@asamalik Could you please let me know when I can examine the result on tiny.distro.builders?

torsava:
@asamalik IIRC it was mentioned on the discussion that this approach might only work for leaf packages, is that true? Because plenty of stuff is going to require python3-pip.

asamalik:
@hroncok It refreshes every ~90 mins, so this should be already live.

@torsava You are correct. This doesn't mangle the depsolver, so if python-pip gets pulled in by other packages, both the proper package and the placeholder show up.

asamalik:
And exactly that happened, yes, python-pip is there twice now. Sorry @hroncok I misunderstood your intention, thought it's just a leaf package...

hroncok:
OK. revert this?

asamalik:
It's being pulled in by `micropipenv`: https://tiny.distro.builders/view-rpm--view-eln--python3-pip.html

Sure, I'll revert it

torsava:
@asamalik So, do we understand it correctly that the only 2 ways to get rid of the unwanted dependency of python3-pip is to:

1. Do a conditional in Fedora, which we can't since the maintainer objects.
2. Wait until April and in the meantime keep the RHEL 9 content set incorrect?

PR #391, opened 2020-11-27
Add `Provides: rubygem(tilt)` to rubygem-tilt placeholder.

voxik:
I hope this syntax is supported.

This is related to #386.

### DIFF
--- a/configs/sst_cs_apps-ruby.yaml
+++ b/configs/sst_cs_apps-ruby.yaml
@@ -15,6 +15,9 @@ data:
   package_placeholders:
     rubygem-tilt:
       description: Generic interface to multiple Ruby template engines
+      provides:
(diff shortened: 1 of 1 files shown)


asamalik:
`package_placeholders` don't reach the depsolver. They're only useful for top-level (leaf) packages. It's literally just an entry on the final list + it pulls all dependencies in.

In other words, Content Resolver always produces a functional and installable set. It can't leave out packages in the middle of a dependency chain.

voxik:
That is bad. Wouldn't it be possible to utilize something like [create-fake-rpm](https://github.com/xsuchy/create-fake-rpm) to provide real placeholder for resolver?

voxik:
This won't be needed in its current form. Closing in favor of #490

PR #414, opened 2020-12-15
texlive-wasy2-ps and texlive-texconfig obsoleted in texlive-2020

ngothan:
tthe workload failed (https://tiny.distro.builders/workload-overview--sst_cs_system_management-text-processing--repository-fedora-eln.html)  because  texlive-wasy2-ps and texlive-texconfig are removed in texlive-2020 upstream.


ngothan:
@asamalik , could you please have a look and merge it?
Thanks

PR #417, opened 2021-01-05
Add tracer to the sst_cs_system_management-other workload

FrostyX:
AFAIK we want to add tracer only because of Cockpit so I am not
creating a separate workload but adding it here.

### DIFF
--- a/configs/sst_cs_system_management-other.yaml
+++ b/configs/sst_cs_system_management-other.yaml
@@ -26,6 +26,7 @@ data:
   - scotch
   - scotch-doc
   - bc
+  - tracer
 
   labels:
   - eln


FrostyX:
I wasn't really sure if I should create a new workload or add it here (or possibly somewhere else). Please somebody point me in the right direction.

FrostyX:
Pinging @dmach as it might interest him.

jwboyer:
If the cockpit team will maintain this, it's correct.  If it's going to be maintained by another team, like the software management team, you should probably add it to their workload instead.

FrostyX:
The cockpit team uses the package but I should be the one maintaining
it. Then I should probably create a workload for myself or Community
Packaging Team (CPT), that I am a part of?

jwboyer:
After some digging, I think it would be best to just add this to configs/sst_cs_system_management-other.yaml  That's the proper owning team and we can get it assigned to you in bugzilla from there.

FrostyX:
Thank you @jwboyer,
moved to `configs/sst_cs_system_management-other.yaml`.

PR #418, opened 2021-01-05
Add modulemd-tools to the sst_cs_system_management-other workload

FrostyX:
Adding the `modulemd-tools` package,
upstream is here https://github.com/rpm-software-management/modulemd-tools

### DIFF
--- a/configs/sst_cs_system_management-other.yaml
+++ b/configs/sst_cs_system_management-other.yaml
@@ -27,6 +27,7 @@ data:
   - scotch-doc
   - bc
   - tracer
+  - modulemd-tools
 
   labels:
   - eln


FrostyX:
Pinging @dmach as it might interest him.

jwboyer:
Do you know which team will maintain this?  It would be better to add it to that team's existing workload so that ownership can be mapped appropriately.

FrostyX:
Thank you for the reply @jwboyer,
per discussion with @dmach and @xsuchy, I am supposed to maintain the package.

So if a single person can be assigned as a maintainer, it should be me. If a team needs to be assigned then I guess it should be Community Packaging Team (CPT), that I am a part of.

jwboyer:
After some digging, I think it would be best to just add this to configs/sst_cs_system_management-other.yaml  That's the proper owning team and we can get it assigned to you in bugzilla from there.

FrostyX:
Thank you @jwboyer,
moved to `configs/sst_cs_system_management-other.yaml`.

PR #421, opened 2021-01-07
Move pinentry to the rest of GPG packages in crypto team

Jakuje:
This mostly summarizes list of dependencies of gnupg2 in RHEL, but also moves `pinentry` package from samba as agreed with Boris and Simo. Adding also @jwboyer as he is marked as a maintainer of the current samba "workload".


jwboyer:
Yes, excellent.

PR #422, opened 2021-01-07
rubygem-bundler is wanted, while it is not wanted as independent package

voxik:
In ELN, we want only the rubygem-bundler which is build from ruby package, while in Fedora, there is also independent rubygem-bundler, which we don't want. It seems that content-resolver cannot make good enough distinction between these two. OTOH, independent rubygem-bundler is hopefully removed already, so we should be able to drop this.

### DIFF
--- a/configs/sst_cs_apps-unwanted-ruby.yaml
+++ b/configs/sst_cs_apps-unwanted-ruby.yaml
@@ -5,9 +5,10 @@ data:
   description: Packages we do not want to ship for Ruby
   maintainer: sst_cs_apps
 
-  unwanted_packages:
+  unwanted_source_packages:
   - rubygem-bundler
-  - rubygem-bundler-doc
+
+  unwanted_packages:
   - rubygem-rspec-its
   - rubygem-rspec-its-doc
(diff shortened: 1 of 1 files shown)


jwboyer:
I'm going to let @asamalik look at this one.  I understand the situation, but not how Content Resolver will deal with it.

voxik:
I think the general problem is that it is not clear if the context is srpm or binary package. And I am affraid that the situation is even more complex, because I guess also virtual provides comes into the play.

asamalik:
`unwanted_packages` is binary RPMs. You can also use `unwanted_source_packages` to mark whole SRPMs as unwanted.

If I understand correctly, you might want to mark the `rubygem-bundler` SRPM as unwanted.

voxik:
How about this? (And sorry for the delay)

asamalik:
That might do the trick, let's see!

voxik:
This looks good (however I probably missed opportunity to check the difference):

https://tiny.distro.builders/view-rpm--view-eln--rubygem-bundler.html

PR #427, opened 2021-01-11
Remove libuv from unwanted due to many bugs and issues (#1895872)

kasicka:
Since there were multiple bugs filed for missing libuv-devel in EL8, I suggest we include libuv and libuv-devel.

https://bugzilla.redhat.com/show_bug.cgi?id=1895872

@voxik @hhorak @jwboyer

### DIFF
--- a/configs/sst_cs_apps-unwanted.yaml
+++ b/configs/sst_cs_apps-unwanted.yaml
@@ -6,7 +6,6 @@ data:
   maintainer: sst_cs_apps
 
   unwanted_packages:
-  - libuv
   - js-uglify
   - web-assets-devel
 


voxik:
Just FTR, in RHEL8, the libuv was initially imported as dependency of Node.js. Nevertheless Node.js bundles libuv in RHEL8 at the end. However, it should be unbundled again in RHEL9. And there are also other users (cmake and ....)

BTW I think it would be nice to elaborate a bit more in the commit message.

asamalik:
@voxik Do you all want to discuss this more, or is this ready to be merged?

voxik:
I don't think this needs discussion, just update the commit message to be more verbose.

jwboyer:
This has been sitting for a week and I'd rather get it merged with a lacking commit message than let it linger longer.

PR #429, opened 2021-01-11
Add CentOS Stream 8: repos.; environment; ported workloads.

james-antill:


jwboyer:
I'm clearly missing something.  Can we elaborate on why we're adding these and what they'll be used for?

james-antill:
There are a few points of interest. At a basic level CentOS 8 is generating composes quite often (significantly more so than CentOS Linux did), and it'd be nice to get that data out to people in some way ... this isn't perfect for that, but it's something and hopefully not a significant burden.

Obviously c8s isn't developing at the same pace as ELN, but it is changing and hopefully will change more as time goes on ... so it'll be nice to have historical data. Also it's somewhat easier to compare c9s to c8s if we have the data from c8s ;).

I'm not sure how many workloads we want though, or if it's worth having non-stream data ... but the initial PR answers that with very small and no.

jwboyer:
Thanks @james-antill .  @asamalik can you merge this whenever you think CR is ready?

asamalik:
Looks like this should be all good, merging!

(Sorry for the delay, I got a bit overloaded.)

PR #434, opened 2021-01-12
Add xapian-core to the list of packages for MDA workload

landgraf:
While cyrus-imapd has built-in search engine it's broken badly since 3.0 release
Xapian is standard search engine for this purposes and fully supported.


landgraf:
Wrong base

PR #445, opened 2021-01-15
Add man-pages

nforro:


jwboyer:
We should probably include man-pages-overrides as well, right?

nforro:
man-pages-overrides is a downstream-only package, it doesn't exist in Fedora.

jwboyer:
OK, I'll merge this now, but you can add packages that are not in Fedora with a package placeholder entry.  You might want to do that with a separate commit.

nforro:
Done in #449.

PR #450, opened 2021-01-19
Omit python-hs-dbus-signature from supported libraries in config file

mulkieran:


mulkieran:
I believe that the arguments in the commit message ought to hold. I have not tested.

jwboyer:
You might want to add this to sst_platform_storage-packages-unwanted.yaml as well.

PR #454, opened 2021-01-21
Change names for CentOS Linux 8/stream.

james-antill:
Virtual provides for compatibility? Bah, who wants those.

### DIFF
--- a/configs/stream-httpd.yaml
+++ b/configs/stream-httpd.yaml
@@ -16,14 +16,14 @@ data:
   - mod_proxy_html
   - mod_ldap
   - mod_session
-  - mod_wsgi
+  - python3-mod_wsgi
   - httpd-devel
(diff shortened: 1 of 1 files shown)


asamalik:
I know... it requires package names by design, because virtual provides can overlap, so this way you can be sure you're monitoring exactly what you intended.

james-antill:
Ahh, to be clear I was complaining that the package python3-mod_wsgi didn't do a virtual provide meaning even "dnf install mod_wsgi" instructions don't work.
I hadn't realized you don't use them anyway, although it makes some sense :) ... Would it be possible to use them if the provide only matches a single package? Eh, I guess that's probably too complicated.

asamalik:
No worries!

Yeah I'm afraid that would be too complicated for a very little benefit.

PR #455, opened 2021-01-22
Pumbers: add zram-generator

lnykryn:


tdawson:
So ... just a short word about this pull request.
zram-generator (whose source rpm is rust-zram-generator) pulls in about 550 rust crates to RHEL9.
It might be too late now, but maybe we should talk about bundling rust-zram-generator, or figure out who's going to maintain those 550 packages in RHEL9.

jwboyer:
There are other ways of dealing with that issue as well.

PR #459, opened 2021-01-26
Add Desktop documentation and branding workload

tpopela:


jwboyer:
Out of sheer curiosity, which repo would you expect those docs packages to ship in.  I'll go first: CRB.  They aren't needed during runtime in production.

tpopela:
> Out of sheer curiosity, which repo would you expect those docs packages to ship in. I'll go first: CRB. They aren't needed during runtime in production.

I don't know. Probably the same as in RHEL 8. The redhat-indexhtml is used by Firefox (in runtime) (at least I think), and gnome-getting-started-docs is the official user documentation that you have for GNOME, that is being displayed in Yelp (Help application) on default installation - so that is definitely used in runtime and in production.

tpopela:
This is from firefox.spec on RHEL:
Requires:       redhat-indexhtml

jwboyer:
Hm.  OK.  I would expect yelp to have Requires on those packages then.  So AppStream it probably is.  Thanks!

PR #461, opened 2021-01-28
printing-stack.yaml: don't require nss-mdns

zdohnal:
nss-mdns can be added during the life cycle, if the solution with systemd-resolved + Avahi doesn't work.

### DIFF
--- a/configs/sst_cs_infra_services-printing-stack.yaml
+++ b/configs/sst_cs_infra_services-printing-stack.yaml
@@ -31,8 +31,6 @@ data:
   - pnm2ppa
   # fonts
   - urw-base35-fonts
-  # resolving .local addresses
-  - nss-mdns
 
 
   labels:


jwboyer:
We aren't using systemd-resolved by default in EL 9.  It will be disabled and customers will have to opt into it.  I don't think you want to drop this right now.

zdohnal:
Aha, ok. Actually, cups functionality which requires nss-mdns is in RHEL for long time (since el7), but no customer actually required it and was satisfied with nss-mdns from epel. So even if systemd-resolved isn't in RHEL 9, it will not be any change from el8. So IMO it is okay to remove it for now and explore possibilities of Avahi-resolved cooperation.

jwboyer:
OK.  Let's drop it then.  We can always add it back if needed.  I just wanted to make sure you were aware that systemd-resovled wasn't the default.

zdohnal:
@jwboyer no problem, thanks for that (it trains my convincing and explaining skills). I actually talked with manager of plumbers team about resolved before that, and he told me they are currently considering whether resolved will be in opt or in main repos, so I will talk to them more later.
But since removing nss-mdns will not cause a regression, I decided to drop it for now.

PR #462, opened 2021-01-28
Add mysql placeholder

hhorak:
MySQL packages are called `communit-mysql` in Fedora, but we want to keep using `mysql` in RHEL as we've always did. Therefore, defining `mysql` as a placeholder for now and specifying all the deps that we'll probably need. Hope the format is correct.


tdawson:
LGTM

PR #471, opened 2021-02-04
Specify mysql-selinux explicitly because conditionals are not handled correctly

hhorak:
package mariadb-server requires (mysql-selinux if selinux-policy-targeted)


asamalik:
They are handled correctly, it's just the selinux-policy-targeted package is not preinstalled.

PR #474, opened 2021-02-08
Fix the list of dependencies for OpenStack

amoralej:
I'm updating this list with the list of:

- Runtime requirements for OpenStack packages
- Packages installed in Dockerfiles for OpenStack containers
- Packages installed in overcloud and ironic-python-agent images

### DIFF
--- a/configs/sst_openstack.yaml
+++ b/configs/sst_openstack.yaml
@@ -6,102 +6,330 @@ data:
   maintainer: sst_openstack
 
   packages:
-  - augeas
-  - autogen
-  - babel
-  - babeltrace
-  - celt051
+  - aide
+  - ansible-freeipa
+  - augeas-libs
+  - bind
+  - biosdevname
+  - boost-atomic
+  - boost-chrono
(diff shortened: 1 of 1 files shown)


configs/sst_openstack.yaml
```diff
@@ -6,102 +6,330 @@ data:
...
+  - pacemaker-remote
+  - parted
+  - pciutils
+  - pcre
   - pcs
+  - perl-Carp
+  - perl-interpreter
+  - perl-libs
+  - pkgconf-pkg-config
+  - platform-python
```
jwboyer, line 193: In the context of ELN, there is no platform-python.  We should probably exclude this one and just focus on the python3 packages listed below to pull in the relevant runtime.

amoralej: > In the context of ELN, there is no platform-python. We should probably exclude this one and just focus on the python3 packages listed below to pull in the relevant runtime.

Right, it makes sense, i'll update my patch.

PR #475, opened 2021-02-09
i18n input-methods: change Japanese IME from ibus-kkc to ibus-anthy

juhp:
https://fedoraproject.org/wiki/Changes/ibus-anthy_for_default_Japanese_IME

### DIFF
--- a/configs/sst_i18n-input-methods.yaml
+++ b/configs/sst_i18n-input-methods.yaml
@@ -13,8 +13,8 @@ data:
   - gtk2-immodule-xim
   - gtk3-immodule-xim
   - ibus
+  - ibus-anthy
   - ibus-hangul
-  - ibus-kkc
   - ibus-libpinyin
   - ibus-libzhuyin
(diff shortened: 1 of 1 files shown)


jwboyer:
@juhp Does this mean we don't want ibus-kkc at all?  If so, it probably needs to be added to the unwanted list.

juhp:
Okay fair point.

juhp:
Could I do that afterwards/later?  Since we don't have a unwanted pipeline yet for i18n.
Or do you prefer to wait for that?

jwboyer:
@asamalik do you have a preference?  I know things are in-flight at the moment.

asamalik:
Let's get this change merged so we can see it on the output. The unwanted package is just a marker for humans to see, it doesn't have an actual functionality in Content Resolver (like it won't affect the dep solving).

juhp:
Ping :)

asamalik:
Sorry!

PR #477, opened 2021-02-10
sst_platform_tools-unwanted.yaml: Update description

fweimer-rh:
The list contains packages the team does not maintain in RHEL 8.

@codonell As discussed off-PR.


codonell:
This looks good to me.

We shouldn't be listing the explicit RHEL versions anyway, and we also have some strong opinions about not including certain packages that are related to our area of expertise (even if we don't own them).

PR #478, opened 2021-02-10
sst_platform_tools-unwanted.yaml: Do not mark nscd as unwanted

fweimer-rh:
nscd should be removed from future ELN, but not the c9s iteration.

@codonell Internal records need updating as well.


jwboyer:
@asamalik is the file/labeling setup on this one going to accomplish what @fweimer-rh wants?  This is brand new as of today, so I have no idea.

asamalik:
yep, that'll do it!

nscd will be marked as unwanted just in eln, the rest remains marked as unwanted in both eln and c9s

PR #479, opened 2021-02-10
Update sst_security_crypto-unwanted.yaml

simo5:
We also want to avoid PyOpenSSL, not because it is a bad package but because upstream want to drop it and provided good alternatives.


jwboyer:
What are those alternatives out of curiosity?

simo5:
python native SSL support for TLS connections and python-cryptography for anything that is not covered there.

jwboyer:
Thanks!

PR #481, opened 2021-02-11
sst_platform_tools-unwanted.yaml: Mark libbsd as unwanted

fweimer-rh:
This continues a tradition I started with Red Hat Enterprise Linux 7. The concerns are pretty much unchanged (the `fgetln` implementation changed radically, but still has problems).

@codonell @JeffreyALaw As discussed internally.


codonell:
This looks good to me. Absolutely something we should not be shipping in the distribution.

jwboyer:
If you can get minizip to drop the dep, you might have a chance this time :)

codonell:
@jwboyer Yes, minizip can be built without libbsd. @fweimer-rh is following up with minizip upstream.

PR #485, opened 2021-02-12
Update unwanted Python 3 packages

hroncok:


encukou:
@hroncok Could you also add `python-pillow`, `python-pretend`, `python-sure`, `python-html5lib`?
(For those, it's reasonably safe to assume they're not included in RHEL9 on purpose. For ~30 others, I'll ask maintainers that are bringing them in.)

hroncok:
What should I put as reason? a general "Python Maint does not want this"?

hroncok:
I'Ve made something up, please review.

hroncok:
I've also invited you to my fork, amend freely.

configs/sst_cs_apps-unwanted-python3.yaml
```diff
@@ -50,14 +54,30 @@ data:
...
+  - python3-pillow-devel
+  - python3-pillow-qt
+  - python3-pillow-tk
+
+  # a test only depndencies, can be replaced by stdlib's unittest.mock if needed
+  # not "enterprise" material
+  - python3-pretend
+  - python3-sure
+
+  # parsing HTML is not something we want to support just for the sake of tests
```
encukou, line 78:
```suggestion
  # The Python standard library already has http.parser;
  # another HTML parsing library is redundant
```

encukou:
Looks good as is, but maybe this would be better for html5lib.

hroncok:
ack, thanks

PR #486, opened 2021-02-12
Update Ruby related config files.

junaruga:
* Add binary RPMs from ruby SRPM except ruby-doc.
* Remove rubygem-bson, rubygem-mongo as we do not ship it.

As seeing the nodejs config file, nodejs-docs was not added to the `<packages>`, I did not add ruby-doc to the `<packages>` too.
https://github.com/minimization/content-resolver-input/blob/master/configs/sst_cs_apps-nodejs.yaml

@voxik review please.

### DIFF
--- a/configs/sst_cs_apps-ruby-db-connectors.yaml
+++ b/configs/sst_cs_apps-ruby-db-connectors.yaml
@@ -6,8 +6,6 @@ data:
   maintainer: sst_cs_apps
 
   packages:
-  - rubygem-bson
-  - rubygem-mongo
   - rubygem-mysql2
   - rubygem-pg
 
--- a/configs/sst_cs_apps-ruby.yaml
+++ b/configs/sst_cs_apps-ruby.yaml
@@ -7,6 +7,26 @@ data:
 
   packages:
   - ruby
+  - ruby-default-gems
+  - ruby-devel
+  - ruby-libs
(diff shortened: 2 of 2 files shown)


configs/sst_cs_apps-ruby-db-connectors.yaml
```diff
@@ -6,8 +6,6 @@ data:
   maintainer: sst_cs_apps
 
   packages:
-  - rubygem-bson
-  - rubygem-mongo
   - rubygem-mysql2
```
voxik, line 9: That is right, we don't want these anymore.

configs/sst_cs_apps-ruby.yaml
```diff
@@ -7,6 +7,26 @@ data:
...
+  - rubygem-psych
+  - rubygem-rake
+  - rubygem-rbs
+  - rubygem-rdoc
+  - rubygem-rexml
+  - rubygem-rss
+  - rubygem-test-unit
+  - rubygem-typeprof
+  - rubygems
+  - rubygems-devel
```
voxik, line 29: If this should list all the binary RPMs instead of SRPM, then this is correct. But I'd rather leave it somebody else to review.

asamalik: yes it's binary RPMs

PR #488, opened 2021-02-15
Update unwanted Ruby packages.

junaruga:
Add rubygem-bson and rubygem-mongo to the unwanted source packages.

The comment style is like https://github.com/minimization/content-resolver-input/blob/253dabe7f9a1f750be471c585d008e4498a778c4/configs/sst_cs_apps-unwanted-perl.yaml#L9

@voxik review please.


voxik:
LGMT

Just FTR, we had discussion if it is worth of the effort listing packages as unwanted, although they are not pulled in by anything and we decided that it would be nice to list them as unwanted just for the purpose of sharing the reason.

Issue #490, opened 2021-02-16
Use RHEL9/CentOS stream dependencies for content resolver


voxik:
It would be nice if content resolver was smart enough to use RHEL9/CentOS stream to resolve its dependencies instead of placeholders. E.g. for git-lfs, there is following placeholder:

https://github.com/minimization/content-resolver-input/blob/253dabe7f9a1f750be471c585d008e4498a778c4/configs/sst_cs_apps-scm.yaml#L18-L28

But it would be much easier, if content resolver knew where to ask RHEL9/CentOS stream for the real values from the package, which is not inherited into ELN.

voxik:
Maybe it would be good start if there was tool, which could based on SRPM/RPMS provide content for the placeholder section, so this would be less manual (error prone) process.

asamalik:
That's already possible, you just need to split the workload into two.

The one for ELN will only use the "eln" label, and the other one only "c9s".

voxik:
> That's already possible, you just need to split the workload into two.

Oh, could you be please more specific? How we could do that with the git-lfs for example?

asamalik:
Copy the existing one to a second file, perhaps `sst_cs_apps-scm-eln.yaml` and only leave the "eln" label (remove the "c9s" one). That will now be just ELN.

In the other one (`sst_cs_apps-scm.yaml`) remove the "eln" label and only keep the "c9s". And move all placeholders into "packages":


```
...
packages:
- git-lfs
...
```

asamalik:
I split these two for example:

[sst_cs_infra_services-varnish-eln.yaml](https://github.com/minimization/content-resolver-input/blob/master/configs/sst_cs_infra_services-varnish-eln.yaml)

[sst_cs_infra_services-varnish.yaml](https://github.com/minimization/content-resolver-input/blob/master/configs/sst_cs_infra_services-varnish.yaml)

voxik:
Thx for explanation. @opohorel could you please try to adjust the git-lfs? Not sure we have it in CentOS streams though, something to explore.

opohorel:
I made a PR with proposed change. Could you please review it and merge if everything is ok. Thank you

voxik:
Where I can see the impact of the PR above? Because I am afraid I don't understand what it actually does. I thought that we will be able to completely get rid of the placeholder ....

voxik:
[...]

PR #491, opened 2021-02-16
split sst_cs_apps-scm workload into two

opohorel:
https://github.com/minimization/content-resolver-input/issues/490#issuecomment-779743596


voxik:
It would definitely deserve better commit message, which would explain what is the intention behind. E.g. that the package inheritance is disabled and we want the content resolver to properly reflect the dependencies used in CentOS

BTW it should not be placeholder anymore, if I understand it correctly.

asamalik:
Yes that'll do what we discussed @voxik

PR #492, opened 2021-02-16
Fix the gnome-desktop workload

tpopela:
Revert the previous change as the gnome-shell-extensions is an SRPM and not RPM


tdawson:
LGTM

PR #493, opened 2021-02-16
Drop libaec from sst_desktop unwanted

tpopela:
It was moved to another subsystem


jwboyer:
Which team is picking it up?

tpopela:
I moved it to sst_cs_system_management per content resolver recommendation. It was on an unwanted list for our sst because none of our package required it.

Issue #424, opened 2021-01-08
buildroot files should not be in this repo


tdawson:
The buildroot files are causing delays and problems for maintainers to manually change their yaml files, and it is only going to get worse.
I've found it takes between 2 and 5 minutes to simply clone the repo.
When manually editing a yaml file, I often have to re-pull and merge because buildroot has changed.
The git repo is currently at 750M.

All of these problems are going to get worse as time progresses.
We need a different way to get the buildroot information to Content Resolver

asamalik:
Wow, that is terrible.

Let's definitely get rid of these and have a different way for the buildroot to get to Content Resolver.

asamalik:
I'll unfortunately need force push to this repo to clean it up.

For archiving purposes, I've created [minimization/content-resolver-input-archive](https://github.com/minimization/content-resolver-input-archive) where I've pushed the current unchanged tree. 

To get this done as quickly as possible, I've also created [minimization/content-resolver-input-additional](https://github.com/minimization/content-resolver-input-additional) which we're gonna use for the large files now.

This repo remains used for configuration as before.

tdawson:
Could you add me, and square1-box as people who can push to the new minimization/content-resolver-input-additional repo.  We can't currently, so I can't change the buildroot-generator to start using that.

tdawson:
I have buildroot-generator setup to use minimization/content-resolver-input-additional but square1-box still cannot push to it.

keszybz:
This can be closed.

PR #499, opened 2021-02-18
change rt-tests to realtime-tests and rt-setup to realtime-setup

jkacur:
rt-tests is now in Fedora and is called realtime-tests
rt-setup is now in Fedora and is called realtime-setup

Change the yaml files to reflect this

### DIFF
--- a/configs/sst_kernel_rats-perf.yaml
+++ b/configs/sst_kernel_rats-perf.yaml
@@ -14,8 +14,8 @@ data:
       - stress-ng
 
   package_placeholders:
-    rt-tests:
-      srpm: rt-tests
+    realtime-tests:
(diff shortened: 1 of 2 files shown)


jwboyer:
@tdawson should the buildroot-eln.yaml file be edited?  I think that's generated automatically.

tdawson:
The buildroot file should NOT be edited.  It is generated automatically.
We have an issue open to move it, and the json files to a different git repo, but due to priorities, we haven't implemented that yet.
https://github.com/minimization/content-resolver-input/issues/424

jkacur:
So, will someone edit the patches to remove the buildroot files, or should
I regenerate the patches?

PR #502, opened 2021-02-18
Drop gnome-shell-extension-horizontal-workspaces from GNOME Classic workload

tpopela:
GNOME 40 uses horizontal workspaces by default, so this is not needed anymore.


jwboyer:
It would be good to put this in the excluded file so it gets removed if we don't want/need it at all.  As it stands right now, it's already built and will just float around unless it's in an excluded section.

tpopela:
> It would be good to put this in the excluded file so it gets removed if we don't want/need it at all. As it stands right now, it's already built and will just float around unless it's in an excluded section.

Makes sense. Thanks!

PR #504, opened 2021-02-19
change rt-tests to realtime-tests and rt-setup to realtime-setup

jkacur:
I regenerated the patches without the buildroot changes.


jwboyer:
Thanks.  I'll merge this as-is.  You can probably just list realtime-tests and realtime-setup as regular packages and drop the placeholder_packages parts.  Those exist in koji now and can be synced.

PR #523, opened 2021-03-09
identity management: make sure ipa-selinux is required in IPA workloads

abbra:
SELinux is expected to be active in IPA workloads but to allow containerized deployments the dependency to ipa-selinux is made optional with the help of rich dependencies:
```
     Requires: ([freeipa]ipa-selinux if selinux-policy-targeted)
```

Make sure that the workload configurations are set to preinstall ipa-selinux where needed. This helps with composes for IPA workloads.

Second commit is for `pki-core` buildroot change where it needs to depend on `ipa-healthcheck` rather than `freeipa-healthcheck`. The latter change for ELN base is tracked with https://github.com/fedora-eln/eln/issues/34


jwboyer:
Thanks!

PR #526, opened 2021-03-15
Remove python-podman-api

jnovy:
The python-podman-api package is no longer required in RHEL9 as it's not part of the container-tools module.


jwboyer:
Probably want to put this into an unwanted section as well.  That way it gets marked for removal.

jnovy:
Can I add the unwanted_packages: section in sst_conatiner_tools.yaml or better to a new file? @jwboyer

jwboyer:
I forget.  @asamalik does CR require separate files or can a single yaml work?

PR #535, opened 2021-03-22
sst_kernel_rats: Separate ELN and CentOS Stream 9 workloads for kerne…

jlelli:
kernel-rt is not going to have ELN packages until PREEMPT_RT is merged
upstream. In the meantime we need to provide packages for CentOS Stream
9.

Separate workloads definitions to make that possible.


asamalik:
Looks good!

PR #538, opened 2021-03-23
Use ursine packages for Maven and Ant

mizdebsk:
Maven and Ant were decided not to be shipped as modules, but ursine packages.


jwboyer:
Can you point me to where that decision was documented (internally is fine)?

PR #540, opened 2021-03-24
connections was renamed to gnome-connections

tpopela:
See https://bodhi.fedoraproject.org/updates/FEDORA-2021-590a47f6b1


jwboyer:
@asamalik @contyk we're going to need a new component and branches for this before we can get a build included and the compose adjusted.

contyk:
I'll comment when it's done in both distros.

contyk:
@jwboyer This should now be done on the infra part, I'll let Adam handle the ownership.

PR #541, opened 2021-03-25
Update samba workloads

abbra:


jwboyer:
Is the AD DC workload only for Fedora?  If we need it for Stream we'd probably need the c9s label.

abbra:
I would love to see c9s with AD DC builds enabled but we have no common agreement on that yet and no actual AD DC builds in C9S/ELN. Right now C9S/ELN follow RHEL packaging layout and features.

I would prefer to merge this as Fedora-only right now and once we have an agreement on C9S, add the label.

PR #542, opened 2021-03-29
sst_kernel_tps-tracing: add opencsd library

rfmvh:
OpenCSD is a library for parsing ARM CoreSight traces in perf.

### DIFF
--- a/configs/sst_kernel_tps-tracing.yaml
+++ b/configs/sst_kernel_tps-tracing.yaml
@@ -15,6 +15,8 @@ data:
   arch_packages:
     aarch64:
       - bpftrace
+      - opencsd
+      - opencsd-devel
     ppc64le:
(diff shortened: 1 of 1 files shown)


configs/sst_kernel_tps-tracing.yaml
```diff
@@ -15,6 +15,8 @@ data:
   arch_packages:
     aarch64:
       - bpftrace
+	  - opencsd
```
asamalik, line 18: This won't work, there needs to be the same indentation (spaces) as for the line above.

asamalik (changes requested):

rfmvh:
OK, changed tabs to spaces. Auto-indentation in my editor used tab for some reason.
 Sorry for not noticing that.

PR #543, opened 2021-03-30
Remove dependency on Sys::Virt and split guestfs-tools from libguestfs

rwmjones:


rwmjones:
FYI guestfs-tools is not actually in Fedora 34 yet, it's been stuck in testing for a while while we fixed some bugs in the upgrade path.  Here is the latest Fedora 34 update: https://bodhi.fedoraproject.org/updates/FEDORA-2021-a24c53741b

rwmjones:
One thing that this PR does not resolve (because I don't know how) is how we will get the new binary packages into RHEL 9.  They are: `guestfs-tools` and `virt-win-reg`.  I'm going to guess that won't just happen automatically?

asamalik:
@contyk Could you please set these two packages up?

rwmjones:
*Just one new package actually.

Do I need to do anything to build the package once dist-git is created?  I was hoping it would be inherited from Fedora 34.

contyk:
Make sure it's in Content Resolver. Besides that you don't really need to do much... or anything.
I'll set up the necessary bits in RHEL and Stream and trigger the initial build for you. I'll comment when it's ready.

contyk:
I set up RHEL and Stream today and attempted to resync guestfs-tools from Fedora into Stream.
However, Stream Koji is currently being borbaded by tagging requests from a different borked service, so I'm not convince that will go through. You or I might need to retrigger it later (Tuesday, potentially).
The easisest way for you to achieve that is to untag the build from f34-cr-eln in Fedora Koji and tag it back in, which should trigger the sync pipeline.

rwmjones:
I don't know what any of that means, but I'll wait until Tuesday when I'm back from holidays, thanks!

rwmjones:
I checked this morning and guestfs-tools dist-git exists, but is empty.

rwmjones:
Do I need to do anything here?  RHEL dist-git still empty.

PR #544, opened 2021-03-30
Another batch of unwanted Python packages

hroncok:
cc @encukou @torsava

### DIFF
--- a/configs/sst_cs_apps-unwanted-python3.yaml
+++ b/configs/sst_cs_apps-unwanted-python3.yaml
@@ -40,6 +40,7 @@ data:
   - python3-ipykernel
   - python3-ipython
   - python3-ipywidgets
+  - python3-jedi
   - python3-jupyter-client
   - python3-jupyter-core
   - python3-matplotlib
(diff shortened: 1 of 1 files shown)


configs/sst_cs_apps-unwanted-python3.yaml
```diff
@@ -74,14 +82,20 @@ data:
...
   # The Python standard library already has html.parser;
   # another HTML parsing library is redundant
   - python3-html5lib
 
-  # This is usually used for %check only and has a huge dep chain
+  # Web servers usually used for %check only, have a huge dep chain
+  - python3-tornado
   - python3-twisted
 
+  # We don't support extra Sphinx themes in RHEL
```
hroncok, line 96:
```suggestion
  # We only support two basic Sphinx themes in RHEL (alabaster and RTD)
```

torsava: Love it. LGTM.

torsava (approved):

PR #545, opened 2021-03-30
redfish-finder: add x86_64 and aarch64 arch exclusivity

theyoyojo:
redfish-finder parses the output of dmidecode and this is only a relevant workflow on x86_64 and aarch64.

Signed-off-by: Joel Savitz <jsavitz@redhat.com>


asamalik:
If you only want it on those two arches, you need to remove it from the packages list. You can leave it empty using `packages: []`

theyoyojo:
Ok will update

theyoyojo:
Fixed

PR #559, opened 2021-04-16
sst_cs_system_management : Some packages in workloads need to be arch specific

ovasik:
When adjusting workloads, forgot to make some packages arch specific, thus make them failing. Fixed with this commit (mostly s390x and ppc64le only packages)


jwboyer:
Thanks @ovasik!

PR #568, opened 2021-04-23
sst_kernel_tps: move perf into its own workload

jeromemarchand:
The perf package currently conflicts with libtraveevent.


jwboyer:
I'm not following why this is needed.  Just to highlight in CR which workload is broken or something?

asamalik:
@jwboyer Content Resolver needs workloads to be installable as a whole, they can't contain conflicts, otherwise the whole thing fails to resolve.

PR #573, opened 2021-05-04
Add glew to desktop unwanted ELN packages

nwnk:
Comment should be self-explanatory.

### DIFF
--- a/configs/sst_desktop-unwanted_eln.yaml
+++ b/configs/sst_desktop-unwanted_eln.yaml
@@ -62,5 +62,8 @@ data:
   - tracker3-miners
   # libhandy1 renamed to libhandy in F34+
   - libhandy1
+  # glew is no longer needed by mesa-demos, and the soname churns too much to be
+  # useful for a long-life product like RHEL.
+  - glew
   labels:
   - eln


jwboyer:
@nwnk I just notice this file doesn't have the c9s label.  That means it won't be reflected in CentOS Stream 9.  If that's expected, then no issues.  If you want it to be though, work with @asamalik to get it squared away in a separate PR.

nwnk:
> @nwnk I just notice this file doesn't have the c9s label. That means it won't be reflected in CentOS Stream 9. If that's expected, then no issues. If you want it to be though, work with @asamalik to get it squared away in a separate PR.

I can't _set_ labels, here, so yeah I'll have to do that.

asamalik:
@nwnk It's the `labels` field in the file:

```
labels:
  - eln
```

If you want it to be in CentOS Stream 9, you need to add `c9s` there, too:

```
labels:
  - eln
  - c9s
```

nwnk:
On reflection this really is ELN-only, it wants excluding because it wants not to be in RHEL. It's fine for CentOS to have a build though since other things in Fedora do use it.

PR #575, opened 2021-05-05
configs: add coreos-installer to sst_edge

runcom:
We need the installer to be part of the system to be used in edge.

cc @nullr0ute

Signed-off-by: Antonio Murdaca <runcom@linux.com>

### DIFF
--- a/configs/sst_edge.yaml
+++ b/configs/sst_edge.yaml
@@ -12,6 +12,7 @@ data:
   - ignition
   - ostree
   - rpm-ostree
+  - coreos-installer
 
   arch_packages:
     x86_64:


jwboyer:
Is this something we'd ship as part of EL, or would it be shipped in another product?

runcom:
> Is this something we'd ship as part of EL, or would it be shipped in another product?

pretty sure as part of EL, I'll sync with Peter to be extra sure but I'm pretty positive there

nullr0ute:
> Is this something we'd ship as part of EL, or would it be shipped in another product?

It would be RHEL or at the very least App Stream

PR #593, opened 2021-06-04
[do not merge yet] Virt fixes

asamalik:
This might need to be reviewed per commit, otherwise it looks messy.

I did two things:

1. Made separate workloads for eln and c9s, and changed all c9s package_placeholders to proper packages
2. This one needs review - included arch packages only in the arches where they are actually available

BTW, there's still a problem, the x86_64 workload currently fails to resolve because `sevctl` doesn't exist. But that's been the case already before this PR, and I believe `sevctl` is a new package that's being added?


asamalik:
@yashkmankad This should fix some of the issues with the workload on c9s. Can you please review this?

yashkmankad:
Hey @asamalik 
Thanks for taking this effort. It makes sense, and this PR LGTM.

I will have @LaneWolf review it as well for the placeholder packages, and to answer your question, yes, sevctl is the new package being added. @connorkuehl is working on that, and can provide and update.

yashkmankad (approved):

connorkuehl:
sevctl is still getting sign-offs and contacts being assigned to it. I don't know what the timeline is like for this to complete, it's my first time doing this :slightly_smiling_face:

PR #594, opened 2021-06-08
edge: Require nss-altfiles

cgwalters:
In practice, rpm-ostree requires this today.  It's not
expressed as a hard dependency because we may do something else
in the future.

### DIFF
--- a/configs/sst_edge.yaml
+++ b/configs/sst_edge.yaml
@@ -12,6 +12,7 @@ data:
   - ignition
   - ostree
   - rpm-ostree
+  - nss-altfiles
   - coreos-installer
 
   arch_packages:


cgwalters:
I was going to add this to coreos, which also depends on `rpm-ostree`, but then I saw:

```
## Do not re-enable in eln without Josh Boyers permission
#  labels:
#  - eln
```

Which came from https://github.com/minimization/content-resolver-input/commit/59027edaddf707ac818d6a86e1ae3e4238ff4a69
which...really seems like it could use some more information in the commit message...

cgwalters:
The motivation here is helping set up RHEL CoreOS 9 builds, but I assume it'd also be useful for Edge/9.

jwboyer:
I don' know the reason for that comment.  Maybe it had something to do with non-RHEL groups adding packages to RHEL through CR.

At any rate, this MR looks fine to me.

cgwalters:
> Maybe it had something to do with non-RHEL groups adding packages to RHEL through CR.

I think of our group as RHEL *and* OpenShift.  We do both.

cgwalters:
Who has merge rights to this?

PR #595, opened 2021-06-09
firewall: Update list of wanted/unwanted packages

n0-1:
Like RHEL8, RHEL9 won't come with legacy variants of arptables, ebtables
or iptables. Since nft variants are provided by iptables-nft, adjust
package lists accordingly.

Drop plain 'iptables' package which no longer exists.

For the sake of completeness, add required -libs packages although they
should come automatically as dependencies.

Add iptables-utils package, it ships useful helper tools.

### DIFF
--- a/configs/sst_networking-firewall-tools.yaml
+++ b/configs/sst_networking-firewall-tools.yaml
@@ -10,17 +10,17 @@ data:
   - c9s
 
   packages:
-  - arptables
-  - arptables-services
   - conntrack-tools
-  - ebtables
-  - ebtables-services
   - firewall-applet
   - firewall-config
   - firewalld
   - ipset
+  - ipset-libs
   - ipset-service
-  - iptables
+  - iptables-libs
   - iptables-nft
+  - iptables-nft-services
   - iptables-services
+  - iptables-utils
   - nftables
+  - python3-nftables
--- a/configs/sst_networking-firewall-unwanted.yaml
+++ b/configs/sst_networking-firewall-unwanted.yaml
@@ -10,8 +10,10 @@ data:
   - c9s
 
   unwanted_packages:
+  - arptables-legacy
   - iptstate
   - bridge-utils
   - ebtables-legacy
   - iptables-legacy
+  - iptables-legacy-devel
   - iptables-legacy-libs


configs/sst_networking-firewall-unwanted.yaml
```diff
@@ -10,8 +10,12 @@ data:
   - c9s
 
   unwanted_packages:
+  - arptables-legacy
+  - arptables-services
```
erig0, line 14: Does the `iptables-nft-services` package have a `Provides: arptables-services` ? If so, then we shouldn't put this in the unwanted list.

configs/sst_networking-firewall-unwanted.yaml
```diff
@@ -10,8 +10,12 @@ data:
...
   - iptstate
   - bridge-utils
   - ebtables-legacy
+  - ebtables-services
```
erig0, line 18: same comment as above.

configs/sst_networking-firewall-tools.yaml
```diff
@@ -10,17 +10,17 @@ data:
...
+  - ipset-libs
   - ipset-service
-  - iptables
+  - iptables-libs
   - iptables-nft
+  - iptables-nft-services
   - iptables-services
+  - iptables-utils
   - nftables
+  - python3-nftables
```
erig0, line 26: This gets pulled in by firewalld, but there is no harm in having it explicit here.

configs/sst_networking-firewall-tools.yaml
```diff
@@ -10,17 +10,17 @@ data:
...
   - firewall-applet
   - firewall-config
   - firewalld
   - ipset
+  - ipset-libs
   - ipset-service
-  - iptables
+  - iptables-libs
   - iptables-nft
+  - iptables-nft-services
```
erig0, line 22: Does this include `iptables-services` ? Or are the separate?

configs/sst_networking-firewall-unwanted.yaml
n0-1: OK, then I'll drop {arp,eb}tables-services from unwanted list.

configs/sst_networking-firewall-tools.yaml
n0-1: They are separate, iptables-services contains iptables and ip6tables services (and is variant-agnostic).

n0-1:
@erig0 I adjusted the patch accordingly, please review again.

configs/sst_networking-firewall-tools.yaml
```diff
@@ -10,17 +10,17 @@ data:
...
   - conntrack-tools
-  - ebtables
-  - ebtables-services
   - firewall-applet
   - firewall-config
   - firewalld
   - ipset
+  - ipset-libs
   - ipset-service
-  - iptables
```
erig0, line 23: I don't think this needs to be removed. I'm sure other SST workloads pull in `iptables`. IIRC, `iptables-nft` has a `Provides: iptables`. That being said, it's better to be explicit with the variant so I think this removal can stay a removal.

erig0 (approved):
LGTM. Thanks Phil!

asamalik:

PR #598, opened 2021-06-15
Adding a builddep to sst_openstack

brenton:
python3-blinker was somehow assigned to sst_openshift even though it has never shipped it nor uses it as best I can tell.  sst_openstack uses python3-oauthlib which has python3-blinker as a builddep.


brenton:
Actually, sst_virtualization is the recommended owner from the content resolver.  I'm just going to report this as unwanted and let them sort it out.

PR #600, opened 2021-06-18
sst_networking: explode tools workload

marceloleitner:
into multiple workloads, so that workload pages have a better, more
focused, list of dependencies for each of these packages.

They are stand-alone packages, and understanding their deps is more
important aggregating them into a single workload.

Signed-off-by: Marcelo Ricardo Leitner <marcelo.leitner@gmail.com>


asamalik:
@marceloleitner Do you want to name each workload something else? Right now they all have the same name, and you won't be easily able to tell which on is which in the UI.

marceloleitner:
Sure. Missed that. Thanks.  v2 in a few.

marceloleitner:
It should be better now, @asamalik .

PR #602, opened 2021-06-21
Drop rubygem-abrt from Ruby package set.

voxik:
There is not ABRT in ELN (AFAIK) therefore there is no use for Ruby ABRT support.


voxik:
@jwboyer please confirm.

jwboyer:
ACK

PR #603, opened 2021-06-21
Added config for automotive packages based on c8s to find dependencies

RajeshKumarSrivastava:
Added a new config with packages for automotive MVP based on c8s. The purpose is to get the dependencies of these packages so that it can be pulled periodically


jwboyer:
There's no labels on this one, and I think we really want it to look like a layered workload in CR.  @asamalik can you advise how to accomplish that?

asamalik:
So this actually has the `centos` label, which currently means it would resolve on top of these two envs:

* [CentOS Stream 8 minimal DNF env](https://tiny.distro.builders/config-env--c8s-minimal-base-dnf.html)
* [CentOS Stream 8 minimal microdnf env](https://tiny.distro.builders/config-env--c8s-minimal-base-microdnf.html)

Is that what you want @RajeshKumarSrivastava?

asamalik:
(sorry I just sent that previous reply sooner by mistake before finishing my answer)

@jwboyer What do you mean by a layered workload? Do you mean so that it's not included in ELN for example? If so, that's not the case here. It would be just a standalone workload, resolved using CentOS Stream 8, and wouldn't be included in any of the views.

jwboyer:
@asamalik by layered I mean that I want it to show up as a set of packages that something that sits on top of EL needs, but not necessarily be considered part of EL.  E.g. if Auto needs libfoobar, but that's not part of EL, then it just shows up as a package that Auto maintains.

I'm trying to avoid the Openstack problem.

RajeshKumarSrivastava:
> So this actually has the `centos` label, which currently means it would resolve on top of these two envs:
> 
> * [CentOS Stream 8 minimal DNF env](https://tiny.distro.builders/config-env--c8s-minimal-base-dnf.html)
> * [CentOS Stream 8 minimal microdnf env](https://tiny.distro.builders/config-env--c8s-minimal-base-microdnf.html)
> 
> Is that what you want @RajeshKumarSrivastava?

I am not sure if those environments restricts the process of resolving dependencies for those packages list. If so, do I need to create a new environment. Pardon me, I cannot make out completely what the environment would be for in this case. can you please help

asamalik:
@jwboyer Got it! The way it's set up now won't make it appear in any of the views (like ELN), so it won't be a problem. And I'll work with them to make it an add-on view so Content Resolver can highlight the extra packages on top of EL.

@RajeshKumarSrivastava Let me merge this so you can see what the output is gonna look like, and l'm more than happy to help to then tweak it and make sure it represents what you need. Let's chat on the side!

PR #615, opened 2021-07-19
Add ctdb

mkosek:
IdM Subsystem was split into a subsystem group - rename respective workloads. Add 1 missing package in Samba workload.


jwboyer:
@asamalik does CR or any of the tooling that uses it know about SSGs?

mkosek:
For the record, in case the the IdM SST-->SSG rename gets controversial, we can solve that in a separate PR and I can strip the first patch out. The omission of ctdb was the primary motivation for this PR.

mkosek:
Thank you. We may end up splitting the IdM workload into it's new subsystems, if we see that current setup is causing issues. (Some of the workloads have ownership across subsystems)

PR #628, opened 2021-07-28
Remove workload for OVAL authoring support

yuumasato:
The `oval-graph` package won't be available for now.

@matejak @dahaic Please take a look too, thanks.


matejak:
LGTM, the workflow needs to be postponed.

PR #629, opened 2021-07-28
Add oval-graph to sst_security_compliance unwanted

yuumasato:
- Add `oval-graph` to list of unwanted packages
- `oval-graph` will be repurposed and its inclusion is postponed


matejak:
@jwboyer you can go ahead and merge this

PR #630, opened 2021-08-04
Add a script for filter packages by sst

hhorak:
It is also possible to use it for finding differences between the compose content
and the content definition config (see the comment in the script).


jwboyer:
Neat!  I'll let @asamalik do the merge, but this looks good to me.

asamalik:
Sure, why not!

PR #633, opened 2021-08-05
Reintroduce cryptsetup subpackages.

oniko:


oniko:
All those packages were in Fedora and RHEL8 (since 8.0). The only exception is cryptseutp-reencrypt that is already in unwanted list

PR #634, opened 2021-08-05
sst_filesystems: Add fmt to unwanted packages

b-ranto:
This package was assigned to  sst_filesystems as the dependency for ceph
in RHEL 9. We managed to remove that dependency so I am marking the
package as unwanted.

Signed-off-by: Boris Ranto <branto@redhat.com>


b-ranto:
@asamalik Can you please review and merge this?

PR #638, opened 2021-08-20
Add bcc-tools to the tracing workload

jeromemarchand:
Bcc-tools package is missing from the compose since we replaced the
hard dependancy on bcc-tools from bcc package by a soft one
(Recommends). Let's add it explicitly to the workload.

Resolves: rhbz#1991897


jwboyer:
We'll need to make adjustments to the compose directly, as CR doesn't influence that.  But this is good to list here anyway.

PR #640, opened 2021-08-24
Make gdb-minimal available on everything but i686.

drsmith2:


jwboyer:
If you need to change something in the composes, you'll need to file a separate ticket.

PR #647, opened 2021-09-08
sst_coreos: correct maintainer

miabbott:
There is no `sst_rhcos`, but instead the label for the CoreOS SST is
just `coreos`.


jwboyer:
There is an sst_rhcos in bugzilla, and Content Resolver uses it to map things to appropriate owners.  I don't think your change is correct.

miabbott:
> There is an sst_rhcos in bugzilla, and Content Resolver uses it to map things to appropriate owners. I don't think your change is correct.

Okey dokey!

PR #648, opened 2021-09-14
sst_networking: Add libqrtr-glib which is required by ModemManager

cathay4t:
The 5G communication require new version of ModemManager which need
libqrtr-glib.


asamalik:

PR #651, opened 2021-10-13
Fix typo in sst_container_tools name

travier:
Rename the SST config file name. I don't know if this impacts other systems elsewhere where the change should be reflected.

If it's too much effort for not much gain, we can also keep things as is.


travier:
Gentle ping for this one @jwboyer

PR #653, opened 2021-10-14
Updates in automotive workloads for neptune app in c8s

RajeshKumarSrivastava:
Add a workload for neptune application
Add another repository with neptune app repo
Updates the existing workload in automotive to support the above

### DIFF
--- a/configs/automotive-c8s-neptune-manifest.yaml
+++ b/configs/automotive-c8s-neptune-manifest.yaml
@@ -1,0 +1,15 @@
+---
+data:
+  description: Automotive Neptune image manifest packages.
+  labels:
+  - automotive-c8s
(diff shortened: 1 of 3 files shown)


RajeshKumarSrivastava:
@asamalik can you please review the changes and merge

Is there a possibility that I can have a base workload like https://github.com/minimization/content-resolver-input/blob/master/configs/automotive-c8s-image-manifest.yaml and another workload as in this PR configs/automotive-c8s-neptune-manifest.yaml which gets resolved on top of the base workload results (so that it has only packages which are not there already in the result of the base workload) and not independently(which otherwise will contain packages resolved already in base workload results) as a separate workload.
Is it possible with views or something since they are related?

asamalik:

PR #654, opened 2021-10-18
Update automotive to the newer config lists.

smooge:
* Move the old automotive items out of the content repository as they
  were for an older
* Add in a pointer to the current neptune repository until we get a
  more permanent place.
* Add in a minimal environment for automotive workloads for base
  enablement
* Add in larger workloads for using neptune3 sample application, qemu,
  podman and some server applications.
* Add in views to give ideas of each.

Signed-off-by: Stephen Smoogen <ssmoogen@redhat.com>


smooge:
This should allow for program to work with these items.

asamalik (approved):

PR #655, opened 2021-10-21
Re enable the disabled automotive workloads

RajeshKumarSrivastava:
These workloads were disabled before. Need to enable them


RajeshKumarSrivastava:
@asamalik can you please merge this. It is urgent

PR #658, opened 2021-10-25
automotive: add osbuild dependencies

juanje:
Added some missing dependencies for building the automotive image with `osbuild


asamalik:

PR #668, opened 2021-11-12
Automotive base 2021112a: finish up items for toolchain

smooge:
This should add in the last of the missing itmes.


michael131468:
This is ready to merge.

PR #669, opened 2021-11-19
Split sst_front_door into installer/image_builder/cockpit

martinpitt:
The previous sst_front_door got changed into an ssg_front_door group
with individual to sst_cockpit, sst_image_builder and sst_installer
teams.


martinpitt:
@teg FYI

PR #672, opened 2021-12-01
Added a CI workflow

regexowl:
Hi @asamalik!

Here's the GitHub Actions workflow for CI for testing the config files. 

The CI checks the config files using `test_config_files.py` from the code repository content-resolver. It then prints the output to a body of PR comment. Which means you should be able to see if the checks passed right away. Here's a demonstration: https://github.com/regexowl/content-resolver-input/pull/50

The `test_config_files.py` runs the `get_configs` function from the `feedback_pipeline.py` to check the configs, so all the code stays in one place.

Let me know what you think :)

### DIFF
--- a/.github/workflows/config-check.yml
+++ b/.github/workflows/config-check.yml
@@ -1,0 +1,52 @@
+name: Check of config files
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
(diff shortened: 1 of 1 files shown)


asamalik:
Very cool!

I already see two broken configs in your example, so I'll go ahead and fix them. :) 

Thanks so much for this.

PR #689, opened 2022-02-07
Drop flex-doc

sgallagher:
This subpackage does not exist.


asamalik:
(the CI error appears the CI is broken, I'll get that fixed)

PR #692, opened 2022-02-14
Include documentation and skopeo

contyk:
This PR includes two changes, one to include skopeo in the set to re-enable osbuild with the updated templates, the other to include documentation in the automotive repository to address functional safety requirements.


asamalik:
You're probably the first using `include-docs`, congratulations!

PR #696, opened 2022-02-21
Include plymouth in automotive definitions where approproitate

ericcurtin:
Include in automotive definitions that contain gdm, mutter.
plymouth also can provide output in the headless case, so
may want to consider adding elsewhere also in future.

Signed-off-by: Eric Curtin <ecurtin@redhat.com>


contyk:
This workload is owned by @telemaco; it could be the right place, it could be not.

telemaco:
I'm ok with the change.

PR #697, opened 2022-02-22
Add Apache Log4j 2 workload

mizdebsk:
Add new Apache Log4j 2 workload.
Also remove log4j and disruptor (dependency of log4j) from unwanted packages.


mizdebsk:
Related: https://issues.redhat.com/browse/CS-1049

PR #699, opened 2022-02-23
Remove tss2

shoracek:
Remove tss2 so that it can be removed from RHEL 9


asamalik:

PR #718, opened 2022-06-24
automotive-repositories: switch to production c9s composes

michael131468:
Switches the CS9 repositories to the production compose as the
development compose is unstable. Currently the annobin packages in the
development compose have bugs that block it from reaching the production
compose and is a case where we would be introducing a buggy state by
using the development compose as an input.
    
Also drops buildroot and epel as we don't want to introduce non c9s
content into the AutoSD distribution by default.
    
Also adds the autosd CBS packages and shift that plus the Automotive CBS
packages to a higher priority layer (would make sense that what they fork
take priority over C9S).

### DIFF
--- a/configs/automotive-repositories-c9s.yaml
+++ b/configs/automotive-repositories-c9s.yaml
@@ -5,29 +5,27 @@ data:
   name: Standard Automotive repositories based on C9S
   description: >
     C9S Repositories for Base Autopackages (BaseOS, AppStream,
-    PowerTools, Extras, EPEL, EPEL-Next, Neptune)
+    PowerTools, CBS AutoSD packages, CBS Automotive packages,
+    Neptune)
   # Contacts: psabata, ssmoogen
   maintainer: automotive-base-enablement
   source:
     repos:
+      cbs-autosd:
+        baseurl: https://buildlogs.centos.org/9-stream/autosd/$basearch/packages-main/
+        priority: 0
+      cbs-automotive:
+        baseurl: https://buildlogs.centos.org/9-stream/automotive/$basearch/packages-main/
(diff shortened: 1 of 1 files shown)


michael131468:
@contyk Hi, can you take a look at this proposal from me? Since the toolchain is using the content-resolver currently as a "desired package set state" form of input, it would be nice to switch to the production c9s compose as a source. Currently the development compose has a buggy annobin package (see BZ) and maybe shows it would be simpler for now to fork of the production compose state. What do you think? (We can also change direction to be not based on the compose but that's going to need some figuring out on how we would like to do that).

I also added the autosd CBS packages to get those covered, and dropped the buildroot/epel repositories since we don't want to have those accidentally in AutoSD.

michael131468:
@contyk If this is okay with you, just let me know and I'll remove the WIP tag from the PR.

michael131468:
Well, two CBS repos now (automotive + autosd). So one added, the other unchanged.

Yeah, maybe it makes sense to separate Neptune/Infotainment somehow so that they can continue using buildroot/epel. I could try do that if you'd like.

I think we have to have the stream repos, our main content is coming from them and we're only cherry-picking from autosd/automotive tagged CBS packages for select things. (The autosd buildlogs repo isn't our autosd repo, it's the cbs autosd tagged packages).

contyk:
I would say let's do this and if it doesn't work out, we can adjust. You're currently the only consumer, as far as I know.

michael131468:
@contyk Let me re-run content-resolver locally and check if this breaks neptune/infotainment atm

michael131468:
Seems to work locally.

PR #728, opened 2022-08-25
Update the list of unwanted Python packages

hroncok:


hroncok:
Followup https://github.com/minimization/content-resolver-input/pull/734

PR #734, opened 2022-10-04
python3-toml is unwnated but was also required

hroncok:
See d70542ac44b9beb5502347550ee7395c7a674afd

## Question

Do I need to create a copy of this file for c9s only that has python3-toml because python3-toml is still wanted there, or do I need to remove to c9s label, or do I not care about c9s here, because packages are never removed from there?


asamalik:
So far we want to keep the c9s as close as possible, so splitting it into a different file (having one for c9s and one for eln) would work the best I think.

hroncok:
Done.

PR #740, opened 2022-10-12
Few font package name updates

pnemade:
1) smc fonts renamed to rit fonts
2) write redhat-fonts binary packages explicitly instead of its source package name

### DIFF
--- a/configs/sst_i18n-extra-fonts-eln.yaml
+++ b/configs/sst_i18n-extra-fonts-eln.yaml
@@ -16,8 +16,7 @@ data:
   - redhat-display-fonts
   - redhat-mono-fonts
   - redhat-text-fonts
-  - smc-rachana-fonts
+  - rit-rachana-fonts
 
   labels:
-  - c9s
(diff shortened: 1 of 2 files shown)


jwboyer:
The smc rename is only in ELN, correct?  If so, we shouldn't change the file that has the c9s label.

asamalik:
If it's only in ELN, you can split the files into two, one having the `c9s`
label and the other the `eln` label.

pnemade:
@asamalik Do you have any example already about splitting file into two?

asamalik:
I do!

https://github.com/minimization/content-resolver-input/blob/main/configs/sst_cs_apps-scm-eln.yaml
https://github.com/minimization/content-resolver-input/blob/main/configs/sst_cs_apps-scm.yaml

There's more. The ELN-specific ones usually end with -eln.

pnemade:
Thanks. I have updated this PR now. Do let me know if it requires any further changes.

PR #742, opened 2022-10-27
Create RHEL packager workload (converted from pablomh workload)

carlwgeorge:
During the early days of CentOS Stream bootstrapping, it is useful for RHEL maintainers to be able to dogfood it (use the distro to develop the distro).  Let's add a specific workload for this to ensure the necessary packages stay compatible.

I spoke with @pablomh about this already over email, and he agreed to allow us to repurpose his single-package workload since it already contained one of the packages we need to include.

cc: @bstinsonmhk


pablomh:
Confirming here what @carlwgeorge said: fine by me :)

PR #743, opened 2022-11-02
Add Jakarta JAXB 2 workload

mkoncek:


mizdebsk:
+1 from me as sst_cs_apps Java PO.
We want JAXB in ELN and CS9.

PR #749, opened 2023-01-05
Remove placeholder for rhel-system-roles

sgallagher:
We now have the correct source name in ELN and this will prevent us from erroneously pulling in pandoc as a BuildRequires.

Signed-off-by: Stephen Gallagher <sgallagh@redhat.com>


sgallagher:
> Do we want to add rhel-system-roles as a proper package? Currently I only see the placeholder here: https://tiny.distro.builders/view-packages--view-eln.html

I don't understand the question. It's (`rhel-system-roles`) been present in the compose for quite some time now.

tdawson:
LGTM
In answer to Adam's question, and to expand on Stephens answer, rhel-system-roles has been in Fedora's koji for almost two years.
https://koji.fedoraproject.org/koji/packageinfo?packageID=33304
So, this pull request really should have been done about two years ago.

tdawson:
That's why there are two files changed.  There is already a configuration with exactly what you are saying, but it is for c9s only.   sst_system_roles-rhel-system-roles.yaml
If you notice, this pull request changes that also, just by adding "eln" to it's views.

asamalik:
Yes. I feel a bit stupid now, you're right.

PR #750, opened 2023-01-13
Add python3-hatchling

yselkowitz:
This is a new dependency of python3-tox, which is required by micropipenv.


hroncok:
I don't understand this PR, sorry.

yselkowitz:
This package was not being kept current despite being a dependency of tox.  We have however discovered the root cause, which is a combination of CR not correctly parsing build deps since the switch of mock in rawhide from chroot to systemd-nspawn and its inability to parse packages with dynamic buildrequires.

PR #758, opened 2023-01-20
Add `iwlax2xx-firmware` package

pablomh:
It's needed for Intel Wi-Fi 6 AX2xx cards functionality.

### DIFF
--- a/configs/sst_kernel_maintainers-kernel-base-eln.yaml
+++ b/configs/sst_kernel_maintainers-kernel-base-eln.yaml
@@ -19,6 +19,7 @@ data:
         - iwl6000g2b-firmware
         - iwl6050-firmware
         - iwl7260-firmware
+        - iwlax2xx-firmware
         - kernel
         - kernel-abi-stablelists
         - kernel-core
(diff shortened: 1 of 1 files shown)


configs/sst_kernel_maintainers-kernel-base.yaml
```diff
@@ -19,6 +19,7 @@ data:
         - iwl6000g2b-firmware
         - iwl6050-firmware
         - iwl7260-firmware
+        - iwlax2xx-firmware
```
sgallagher, line 22: If this package is only intended for ELN/RHEL 10, then please copy this workload to a new file and make the change there. Then set the `labels:` appropriately.

If this package is being added also to c9s, then the current patch is fine.

sgallagher (changes requested):

configs/sst_kernel_maintainers-kernel-base.yaml
pablomh: Let me investigate the CS9 part and I'll get back to you and act accordingly.

pablomh: I've checked that the required firmware is shipped in C9S by the `iwl7260-firmware` package. Please let me know if the split I've done is correct.

PR #759, opened 2023-01-20
Add `fwupd-efi` package

pablomh:
The package is needed on EFI systems in recent `fwupd` versions and is not defined as a dependency of `fwupd` as explained in [Bug 1993508 - /usr/libexec/fwupd/efi/fwupdx64.efi cannot be found](https://bugzilla.redhat.com/show_bug.cgi?id=1993508) so it needs to be added explicitely.


configs/sst_desktop_firmware_bootloaders-fwupd.yaml
```diff
@@ -7,6 +7,7 @@ data:
 
   packages:
   - fwupd
+  - fwupd-efi
```
sgallagher, line 10: If this package is only intended for ELN/RHEL 10, then please copy this workload to a new file and make the change there. Then set the `labels:` appropriately.

If this package is being added also to c9s, then the current patch is fine.

sgallagher (changes requested):

configs/sst_desktop_firmware_bootloaders-fwupd.yaml
pablomh: Let me investigate the C9S part and I'll get back to you and act accordingly.

sgallagher (approved):

PR #779, opened 2023-03-06
Add the first printer applications

zdohnal:


jwboyer:
If this is only intended for ELN, you probably want to create an ELN specific config file for these additions.  Otherwise they'll show up in the c9s content set as well.

zdohnal:
Hi @jwboyer ,

thank you for the review! I would like to add those applications into ELN and then to the future CentOS Stream 10. Are the current ELN related configs going to be automatically cloned into c10s or manual change is required?

zdohnal:
Ok, hard resetting the branch closes the PR...

PR #780, opened 2023-03-06
Add lprint and ipp-usb to ELN and C10S

zdohnal:
lprint provides support for common label printers, which don't support any driverless standards.

ipp-usb enables driverless printing for USB printers, which have support for IPP-over-USB protocol.


asamalik:
btw we don't have a dedicated c10s, eln is that now, but it doesn't hurt anything to have the label there... when we branch c10s in the future we'll just add those labels everywhere anyway

zdohnal:
@asamalik that's good to hear, thanks!

PR #790, opened 2023-04-05
Remove gutenprint-libs-ui from C10S

zdohnal:
gtk2 is not in C10S, remove the dependent gutenprint subpackage.


zdohnal:
Thank you, Adam!

PR #793, opened 2023-04-16
Remove automotive-sig from automotive-repositories

michael131468:
The AutoSD project is using the content-resolver output to define the product definition (contents) of the AutoSD repositories (aka product-build). This commit strips the automotive-sig repositories and packages from the content-resolver workload/repositories because they're not considered part of the AutoSD product and are dirtying the product definition (and muddying the line between automotive-sig and AutoSD).

This is needed to create a cleaner divide between the AutoSD project and the automotive-sig project. Later if these packages are needed in AutoSD, they need to be properly promoted to the autosd tag in order to add them here.

### DIFF
--- a/configs/automotive-environment.yaml
+++ b/configs/automotive-environment.yaml
@@ -20,7 +20,6 @@ data:
   # in-vehicle set without duplicating the base required minimum.  It
   # doesn't mean their reference images need to be ostree-based.
   packages:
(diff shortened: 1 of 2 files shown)


michael131468:
@pypingou ^ Look okay to you?

pypingou:
it does, :+1: for me

PR #797, opened 2023-04-18
automotive-environment: add linux-firmware

michael131468:
The removal of linux-firmware-automotive leaves a gap that needs to be filled. Put linux-firmware in its place.


michael131468:
@pypingou I was a bit hasty simply removing linux-firmware-automotive. We should re-add linux-firmware until linux-firmware-automotive makes its debut in AutoSD.

pypingou (approved):
LGTM

PR #801, opened 2023-05-03
Placeholder more HA packages

yselkowitz:
These packages are built with bundled dependencies in RHEL, and using the Rawhide versions in ELN is pulling in a lot of unwanted Python and Ruby packages.

- Placeholder pcs
- Placeholder more fence-agents


jwboyer:
@feist seem accurate?

feist:
@jwboyer Yes, that is accurate, we bundle most of the library dependencies for RHEL HA in RHEL (but not in Fedora).  I'm not quite sure how to read the commits, but let me know if you have any additional questions there.

PR #805, opened 2023-05-04
Replace Gedit with GNOME Text Editor

tpopela:
Follow what we've done in Fedora Workstation as well, see https://issues.redhat.com/browse/DESKTOP-737 and
https://pagure.io/fedora-workstation/issue/248

### DIFF
--- a/configs/sst_desktop-gnome-text-editor.yaml
+++ b/configs/sst_desktop-gnome-text-editor.yaml
@@ -1,13 +1,12 @@
 document: feedback-pipeline-workload
 version: 1
 data:
-  name: Gedit
+  name: GNOME Text Editor
   description: Text editor for Workstation
   maintainer: sst_desktop
 
   packages:
-  - gedit
-  - gedit-plugins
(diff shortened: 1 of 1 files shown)


configs/sst_desktop-gnome-text-editor.yaml
```diff
@@ -0,0 +1,12 @@
...
+data:
+  name: GNOME Text Editor
+  description: Text editor for Workstation
+  maintainer: sst_desktop
+
+  packages:
+  - gnome-text-editor
+
+  labels:
+  - c9s
```
yselkowitz, line 12: Shouldn't this have been `eln`?

jwboyer: Yes.  @tpopela want to fix this up?

tpopela:
Ah! Thank you @yselkowitz for spotting it! I've opened https://github.com/minimization/content-resolver-input/pull/809 to fix it.

PR #812, opened 2023-05-09
Move kpatch from sst_kernel_ft to sst_kernel_security

joe-lawrence:
A new Kernel Security SST was formed out of the Kernel FT SST.  Update the kpatch content accordingly.


joe-lawrence:
> You probably want to adjust the `maintainer: sst_kernel_ft` line in the file as well.

D'oh, I had left that update staged, but didn't commit.  Thanks for checking that!

PR #820, opened 2023-05-16
Temporarily disable c9s environment

sgallagher:
Currently, the turnaround time to process through both c9s and ELN is over ten hours, which is making it difficult to see the results of changes made to reduce dependencies as we approach the initial import for RHEL 10. Dropping the c9s workloads will reduce this time by at least 33% (and more as ELN gets smaller).

We are not aware of anyone relying on the c9s results at the moment, but if this turns out to be problematic, it's easily reverted.


jwboyer:
Is there no way to make c9s update at a period rate?  Say, once a week?  Teams are still adding packages to their c9s workloads and I do use the results from time to time.  I can live without it if we need to, but it seems like we could change a cronjob or something instead of disabling it entirely.

sgallagher:
Unfortunately, no. The way Content Resolver is set up right now, it's just a bash loop that processes everything in the configuration each time the previous run finishes.

This is very much a stopgap solution while we explore other ways to optimize its processing.

asamalik:
Note this will remove every single c9s workload from the results and all content from the c9s package view.

tdawson:
I did take a backup of the content resolver pages on May 15,  before we turned off c9s.
I have deployed this internally and will send you a link.
If there is anyone who wants this snapshot deployed in a public area, please let me know.

PR #821, opened 2023-05-17
Update sst_platform_tools-misc.yaml removing tbb

nickclifton:
Removes the TBB package from the sst_platform_tools-misc collection as it is only needed when developing using certain experimental features of C++-17


jwboyer:
This removes it from both the ELN and C9S workloads.  This is already shipped in C9S and can't be dropped from the repos, so it would be better to create an unwanted list for ELN and put the TBB packages there.

PR #822, opened 2023-05-17
Exclude the TBB components from Platform Tools for ELN

nickclifton:
Removes the TBB packages from the sst_platform_tools-misc config for ELN


yselkowitz:
@nickclifton this has caused tbb to be listed as both required and unwanted in ELN; we'll need to split your misc workload to drop it from ELN there while leaving it in c9s.

Also, what is the plan for dyninst which currently requires tbb?

PR #824, opened 2023-05-18
gcc no longer requires these packages for RHEL 10

markobrien1:


configs/sst_platform_tools-compat.yaml
```diff
@@ -1,41 +0,0 @@
...
-    the GNU Toolchain.
-  maintainer: sst_platform_tools
-  
-  packages: []
-
-  package_placeholders:
-    # RHEL 8 bumped SONAME of libgfortran.so.  Support RHEL 7 apps.
-    compat-libgfortran-48:
-      description: RHEL-only compat package.
-      srpm: compat-libgfortran-48
```
jwboyer, line 16: This package is still needed in RHEL 9, correct?  If so, you don't want to delete this file.  You just want to remove the `eln` label from it.

markobrien1: That is correct. I have restored the file and just removed the eln label. Thanks @jwboyer

PR #825, opened 2023-05-23
Fix python-tornado status

yselkowitz:
python-tornado was re-added to RHEL 9 post-GA for the Keylime workload.


jwboyer:
Should probably be added to the sst_security_special_projects-keylime.yaml file at the same time.

yselkowitz:
It already is: https://github.com/minimization/content-resolver-input/blob/main/configs/sst_security_special_projects-keylime.yaml#L15

jwboyer:
My grep-fu failed me.

PR #830, opened 2023-05-28
Update Qt6 workloads in ELN

yselkowitz:
Add new Qt6 components in 6.5.1, and limit the Qt6 ELN Extras workload to packages not in ELN proper.

/cc @tdawson @tpopela


tpopela:
I've talked with Jan Grulich and this change looks good to us.

PR #835, opened 2023-06-05
Allow the filtering tool to list unwanted packages if needed

hhorak:
...just seems handy


bstinsonmhk:
@asamalik this looks good to me, but what do you think?

hhorak:
I've added also `--label`, which I'm now using for getting packages for `eln` and `c9s` only to compare what do we have different.

PR #836, opened 2023-06-06
maven-mapping is no longer unwanted by sst_cs_apps

mizdebsk:
maven-mapping is a build-dependency of bnd-maven-plugin, which is needed in RHEL 10
https://src.fedoraproject.org/rpms/aqute-bnd/pull-request/19


yselkowitz:
Nack, see https://src.fedoraproject.org/rpms/aqute-bnd/pull-request/20

jwboyer:
Well, now I'm confused.  Fedora PR 20 says it's unwanted, PR 19 says it's wanted.  I'll let the two of you sort it out.

PR #837, opened 2023-06-06
Remove libdb from supported packages

fila43:
SInce libdb is deprecated and probably will be unwanted, so we should highlight it.
@hhorak PTAL


hhorak:
Yes, it makes sense, thansk.
My nitpicking alter-ego would only suggest linking to the fedora change to provide more context in the commit message, but it's possible to be merged as it is.

PR #839, opened 2023-06-06
Add python3-installer and python3-pyproject-hooks to Python CRB workload

hroncok:
Depends-on: https://src.fedoraproject.org/rpms/python-pyproject-hooks/pull-request/2

cc @torsava

### DIFF
--- a/configs/sst_cs_apps-python-crb.yaml
+++ b/configs/sst_cs_apps-python-crb.yaml
@@ -24,6 +24,11 @@ data:
   - python3-psutil-tests
   - py3c-devel
   - py3c-doc
+  # the following packages are anticipated to be used by pyproject-rpm-macros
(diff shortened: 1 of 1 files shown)


hroncok:
cc @torsava

jwboyer:
Thank you for the do not merge notice.  I'll wait until that's removed.  This is where Gitlab's Draft feature would help :)

jwboyer:
> > This is where Gitlab's Draft feature would help :)
> 
> This PR is actually marked as Draft (although I mistakenly not opened it as such immediately, but only changed it right after opening it).

Yep, caught that immediately after I hit the comment button.  Thanks

yselkowitz:
No objections once pyproject-hooks deps are pruned through that PR.

PR #842, opened 2023-06-08
Python: The flit and poetry tools are not wanted, use -core instead

hroncok:
cc @torsava


bstinsonmhk:
@hroncok do you need a review from @torsava or are you good to merge this now?

hroncok:
Merge please. The cc was just a heads up.

PR #845, opened 2023-06-08
Update program workloads

bstinsonmhk:
Take assignment for a few things and reflect some font changes made in Rawhide that affect the lorax workload.


jwboyer:
:cry: goodbye Program workloads.  You are in better hands with @bstinsonmhk but I will miss you.

As a separate item, we should probably discuss the lorax workload and see if it's still needed in light of the current Anaconda and Image Builder work.

PR #850, opened 2023-06-12
Remove pmem libraries

JeffMoyer:
Intel is dropping support for these libraries.  It's not clear whether other maintainers will step up to develop/maintain them.

### DIFF
--- a/configs/sst_storage_io-packages-unwanted.yaml
+++ b/configs/sst_storage_io-packages-unwanted.yaml
@@ -9,6 +9,11 @@ data:
   - libhbaapi
   - libhbalinux
 
+  unwanted_source_packages:
+  - nvml
+  - libpmemobj-cpp
+  - pmdk-convert
+
   labels:
   - eln
   - c9s


yselkowitz:
Some of these (binary) packages are listed as required, particularly in `configs/sst_storage_io-packages.yaml` and `configs/sst_virtualization-all-eln.yaml`.  Please be sure to remove them from ELN workloads, which may require splitting out c9s configs where they'll remain in that version.

JeffMoyer:
> Some of these (binary) packages are listed as required, particularly in `configs/sst_storage_io-packages.yaml` and `configs/sst_virtualization-all-eln.yaml`. Please be sure to remove them from ELN workloads, which may require splitting out c9s configs where they'll remain in that version.

I don't know how to do the splitting that you suggest.  Is there documentation I can read?

yselkowitz:
@JeffMoyer this is just for ELN (RHEL 10+), right?  These will continue to be supported in c9s/RHEL9?

JeffMoyer:
Yes, that correct.  Should I have mentioned that in the commit description?  Sorry for the confusion.

yselkowitz:
The file to which you added these as unwanted is tagged both `c9s` and `eln`.  So both the (c9s) inclusions and (eln) exclusions needed to be split out.

yselkowitz:
@JeffMoyer how about https://github.com/minimization/content-resolver-input/compare/main...yselkowitz:content-resolver-input:pmem-removal ?

JeffMoyer:
> @JeffMoyer how about [main...yselkowitz:content-resolver-input:pmem-removal](https://github.com/minimization/content-resolver-input/compare/main...yselkowitz:content-resolver-input:pmem-removal) ?

Wow, thanks for doing that!  So, if I understand this correctly, you've created a new workload called sst_storage_io-pmem, and listed all of the binary packages for c9s.  Then, you removed the source packages for that workload in eln, and removed the binary packages from the virt eln workload definition.

If that's right, then I agree with the changes.

Again, thanks a lot for your help!

PR #852, opened 2023-06-13
Fixup the storage_io unwanted packages

bstinsonmhk:
@yselkowitz this appears to be the only difference. Did I get everything this time?


yselkowitz:
@bstinsonmhk that should finish the fixup.

PR #857, opened 2023-06-15
More RHEL 10 desktop removals

tpopela:
Drop evolution-mapi, gnome-photos, inkscape, libgdata and mark enchant, dbus-glib as unwanted.


yselkowitz:
LGTM

PR #867, opened 2023-06-20
Remove pigz

prarit:
pigz was used to compress and decompress gzip'd kernel modules.  It's no longer used AFAICT anywhere in anaconda or for kernel module compression.  It can be dropped as a dependency, which should eventually lead to its removal from RHEL and Fedora.  It is also worth noting that pigz hasn't been updated in YEARS and also points to the packages' lack of use.

### DIFF
--- a/configs/automotive-workload-off.yaml
+++ b/configs/automotive-workload-off.yaml
@@ -86,7 +86,6 @@ data:
     - perl-Carp
     - perl-devel
     - perl-interpreter
-    - pigz
     - pkgconf
     - polkit
     - python3
--- a/configs/sst_arch_hw.yaml
+++ b/configs/sst_arch_hw.yaml
@@ -26,7 +26,6 @@ data:
   - alsa-utils
   - speexdsp
   - hwloc
-  - pigz
   - ethtool
 
   arch_packages:
--- a/configs/sst_program-lorax-template-eln.yaml
+++ b/configs/sst_program-lorax-template-eln.yaml
@@ -16,7 +16,6 @@ data:
   - dnf
   - rpm-ostree
   - ostree
-  - pigz
   - kernel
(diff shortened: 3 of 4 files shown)


configs/automotive-workload-off.yaml
```diff
@@ -86,7 +86,6 @@ data:
     - perl-Carp
     - perl-devel
     - perl-interpreter
-    - pigz
```
jwboyer, line 89: Did you confirm this with the auto SIG?

yselkowitz:
https://tiny.distro.builders/view-rpm--view-eln--pigz.html shows some runtime dependencies remain.

bstinsonmhk:
@prarit do you want to do the c9s/eln split here, or should we rethink this PR?

prarit:
> @prarit do you want to do the c9s/eln split here, or should we rethink this PR?

I'm chasing down answers from SIG, etc., as @jwboyer requested.

lnykryn:
We use it in dracut when it is installed. https://bugzilla.redhat.com/show_bug.cgi?id=1549251

bstinsonmhk:
@prarit any updates here? 

@lnykryn is pigz support something we want to keep around for the initrd?

lnykryn:
I don't have any strong opinions. It is not a hard dependency, but based on the report it significantly increases the creation speed of initrds.

yselkowitz:
@prarit any updates here?

yselkowitz:
@prarit @dledford any updates here?

prarit:
@yselkowitz unfortunately this cannot happen.  The package is needed by a few other things including anaconda.  The package will be reassigned shortly to someone who can spend time in decoupling pigz from the rest of Fedora (and the next version of RHEL).

PR #868, opened 2023-06-20
Add pcre and pcre-devel as unwanted

ljavorsk:
Pcre project has been deprecated and unsupported by upstream since Fedora 38 Fedora Change: https://fedoraproject.org/wiki/Changes/PcreDeprecation

### DIFF
--- a/configs/sst_cs_apps-apps-unsorted-c9s.yaml
+++ b/configs/sst_cs_apps-apps-unsorted-c9s.yaml
@@ -1,0 +1,23 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Other packages from SST CS Aps for c9s
+  description: Unsorted tools owned by CS Apps SST
+  maintainer: sst_cs_apps
+
+  packages:
+  - less
+  - time
+  - pcre2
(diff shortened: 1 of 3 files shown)


configs/sst_cs_apps-apps-unsorted.yaml
```diff
@@ -10,8 +10,6 @@ data:
   - time
   - pcre2
   - pcre2-devel
-  - pcre
```
jwboyer, line 13: This is included in CentOS Stream 9 already.  You can't drop it from this file with the `c9s` label included.  You likely want to create a new eln specific workload and unwanted file instead.

ljavorsk:
Did I fix it correctly @jwboyer ?

yselkowitz:
No, because it is still listed as required in `sst_cs_apps-apps-unsorted.yaml`.  In addition to the eln unwanted file, the c9s and eln configs for apps-unsorted need to be split into separate files, with the pcre/pcre-devel packages removed from the eln config.

ljavorsk:
I've split the files, can you check them now, please?

Sorry for the confusion, it's my first time with content resolver

PR #878, opened 2023-06-22
Explain why nasm and yasm are unwanted

nickclifton:
This patch adds a comment to the ssg_platform_tools-unwanted file explaining why the nasm and yasm tools are needed, and why we would like to drop them.

### DIFF
--- a/configs/ssg_platform_tools-unwanted.yaml
+++ b/configs/ssg_platform_tools-unwanted.yaml
@@ -23,6 +23,8 @@ data:
   - dpkg
   - dpkg-devel
   - dselect
+  # nasm and yasm are required by firefox.  If we could remove this
+  # dependency then we could probably drop them entirely.
   - nasm
   - nasm-rdoff
   - yasm


yselkowitz:
There is a lot more than that currently using these:

https://tiny.distro.builders/view-rpm--view-eln--nasm.html
https://tiny.distro.builders/view-rpm--view-eln--yasm.html

nasm is also being specified for CRB, while yasm is buildroot only (as of 9).  Are you trying to just keep these from being shipped (iow keep them buildroot-only) or are you trying to drop them from the buildroot too?

jwboyer:
It is important to remember that we can change what is included in CRB (or any other repo).  Current existence in a repo isn't a rationale that forces us to retain a package.

nickclifton:
Ah - I had forgotten about the other packages that need nasm and yasm.  Plus I did not know that tiny.distro.builders could show build dependencies.  I will withdraw this pull request and submit another one with a more complete comment.

As for withdrawing the packages from the CRB, whilst that would be nice, it is not really the aim here.  What we would like is to pass ownership of nasm and yasm over to one of the packages that has them as a build dependency.  Then they can decide when/if/how they want to remove that dependency and move over to better assembler (binutils or llvm).

jwboyer:
I already merged it.

yselkowitz:
In the context of CR, "unwanted" is supposed to mean that it does not belong in RHEL at all and should be removed entirely if somehow pulled in.  That does not sound like the case here.

PR #879, opened 2023-06-22
Split off sst_desktop_platform_technologies part of GNOME Desktop workload

tpopela:


configs/sst_desktop_platform_technologies-gnome-desktop-eln.yaml
```diff
@@ -0,0 +1,31 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: GNOME Desktop
+  description: GNOME Desktop core packages
```
yselkowitz, line 5: Is having two workloads in ELN with the same name/description going to cause any problems?

tpopela: https://github.com/minimization/content-resolver/blob/master/config_specs/workload.yaml#L20 doesn't mention that it's supposed to be unique.

jwboyer: We're going to merge it and find out one way or another.  Can fix it up if needed.

PR #881, opened 2023-06-23
More comments about nasm and yasm

nickclifton:
This patch improves the comments describing the nasm and yasm packages in the platform tools unwanted list.

### DIFF
--- a/configs/ssg_platform_tools-unwanted.yaml
+++ b/configs/ssg_platform_tools-unwanted.yaml
@@ -23,10 +23,17 @@ data:
   - dpkg
   - dpkg-devel
   - dselect
-  # nasm and yasm are required by firefox.  If we could remove this
-  # dependency then we could probably drop them entirely.
+  # Nasm is a build requirement for: dav1d, libavif, libjpeg-turbo,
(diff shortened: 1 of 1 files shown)


yselkowitz:
In the context of CR, "unwanted" is supposed to mean that the package does not belong in RHEL at all, and should be removed if/when pulled in, not just that a particular SST is dropping it and putting it "up for grabs".  Unfortunately, not all currently "unwanted" packages conform to that definition.  I appreciate the clarification here but I'm not sure how to handle this.

jwboyer:
If we don't mark it unwanted, it gives no indication to anyone else that the owner doesn't think it should be in the OS.  I think we leave it marked as unwanted and if the current owner can't remove the dependencies or find another owning team, they are left maintaining an unwanted package.  That's just how it goes sometimes.

yselkowitz:
I'll soon be analyzing all the "unwanted" packages to determine which kind of "unwanted" it actually is.  At a minimum, unwanted packages should not also be Required (listed as wanted by another workload) nor explicitly listed to ship in CRB (as nasm currently is), because that would simply be contradictory.

PR #884, opened 2023-06-26
Remove compat-openssl11

neverpanic:
We do not plan on supporting OpenSSL 1.1 in 10, so drop it from the package set. There should be no users of this in the distribution since we did not ship the development package.

### DIFF
--- a/configs/sst_security_crypto-openssl-c9s.yaml
+++ b/configs/sst_security_crypto-openssl-c9s.yaml
@@ -9,6 +9,7 @@ data:
 
   packages:
     - openssl
+    - compat-openssl11
 
   labels:
-    - eln
+    - c9s


neverpanic:
Actually, this is probably not correct, since it drops `compat-openssl11` from the `c9s` label, but we'd only want to drop it from the `eln` label. Any specific guidance on how to do that?

yselkowitz:
Split the configs into separate files, one for c9s and one for eln.

neverpanic:
> Split the configs into separate files, one for c9s and one for eln.

Thanks for the pointer. Done.

configs/sst_security_crypto-openssl-c9s.yaml
```diff
@@ -0,0 +1,15 @@
...
+  description: >-
+    General purpose cryptographic library and toolkit for TLS and SSL protocols.
+  maintainer: sst_security_crypto
+
+  packages:
+    - openssl
+    - compat-openssl11
+
+labels:
+  - c9s
```
yselkowitz, line 15: Indenting needs to be fixed:
```suggestion
  labels:
    - c9s
```

yselkowitz (changes requested):

configs/sst_security_crypto-openssl-c9s.yaml
neverpanic: Yeah, I'm looking at that, and looking at a PR for content-resolver to improve the error message.

yselkowitz (approved):

neverpanic:
See also https://github.com/minimization/content-resolver/pull/50

PR #889, opened 2023-06-26
networking: remove libteam

marceloleitner:
libteam was deprecated on RHEL9 with the intention on removing it on RHEL10. So lets just do it!

References:
https://bugzilla.redhat.com/show_bug.cgi?id=1935544 https://bugzilla.redhat.com/show_bug.cgi?id=1937155 https://gitlab.com/redhat/centos-stream/rpms/libteam/-/merge_requests/2/diffs

@thom311


configs/sst_networking-management-server.yaml
```diff
@@ -19,10 +19,6 @@ data:
   - NetworkManager-ovs
   - nmstate-plugin-ovsdb
 
-  # Team
-  - NetworkManager-team
```
bstinsonmhk, line 23: Since these are still in Stream 9 you probably want to copy the current state into a sst_networking-management-server-c9s.yaml file and move the c9s label over there

marceloleitner: Oh. I thought CS9 was snapshoted a long time ago.
Ok, I'll update the PR. Thx.

marceloleitner:
@bstinsonmhk apparently the merge above was done the other way around? :-)

PR #892, opened 2023-06-27
Add rhc-worker-bash in configs

r0x0d:
Patch to introduce the rhc-worker-bash in BaseOS/AppStream.

### DIFF
--- a/configs/sst_conversions.yaml
+++ b/configs/sst_conversions.yaml
@@ -1,0 +1,30 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: RHC Worker Bash
+  description: Remote Host Configuration (rhc) worker for executing bash scripts on hosts managed by Red Hat Insights.
+  maintainer: sst_conversions
+
+  packages:
+    - rhc-worker-bash
+  
+  labels:     
+  - c9s # TODO(r0x0d): Verify if this is the correct label for RHEL 7
+
+  arch_packages:    
+    x86_64:
+    - rhc-worker-bash
+
+  # Add packages to the workload that don't exist (yet) in the repositories.
+  package_placeholders:
+    - srpm_name: rhc-worker-bash  # SRPM name (mandatory)
+      build_dependencies:   # build dependencies (RPM names) of this SRPM (optional)
+        - golang
+      limit_arches:         # can be left empty for all arches (optional)
+        - x86_64
+      rpms:                 # List of the binary RPMs (mandatory)
+        - rpm_name: rhc-worker-bash   # RPM name (mandatory)
+          description: Remote Host Configuration (rhc) worker for executing bash scripts on hosts managed by Red Hat Insights.
+          dependencies:     # runtime dependencies (RPM names) of this RPM (optional)
+            - rhc
+          limit_arches: []  # can be left empty for all arches (optional)


configs/sst_conversions.yaml
```diff
@@ -0,0 +1,30 @@
...
+    x86_64:
+    - rhc-worker-bash
+
+  # Add packages to the workload that don't exist (yet) in the repositories.
+  package_placeholders:
+    - srpm_name: rhc-worker-bash  # SRPM name (mandatory)
+      build_dependencies:   # build dependencies (RPM names) of this SRPM (optional)
+        - golang
+      limit_arches:         # can be left empty for all arches (optional)
+        - x86_64
```
jwboyer, line 24: Why is this x86_64 only?  The rhc package is offered on all architectures.

configs/sst_conversions.yaml
```diff
@@ -0,0 +1,30 @@
...
+data:
+  name: RHC Worker Bash
+  description: Remote Host Configuration (rhc) worker for executing bash scripts on hosts managed by Red Hat Insights.
+  maintainer: sst_conversions
+
+  packages:
+    - rhc-worker-bash
+  
+  labels:     
+  - c9s # TODO(r0x0d): Verify if this is the correct label for RHEL 7
```
jwboyer, line 12: RHEL 7 doesn't have a CentOS Stream equivalent.  This is the correct label for anything going into CentOS Stream/RHEL 9.

bstinsonmhk, line 12: Is this only for EL7, or do you expect this to come in to EL8/EL9/EL10 too?

configs/sst_conversions.yaml
r0x0d: Not exactly sure if the worker will have support for all other architectures. For now, I added x86_64 as a placeholder. Will confirm with other people to make sure what architectures we will be supporting.

configs/sst_conversions.yaml
r0x0d: Only for EL7 for now.

r0x0d: Oh right, I just want to deploy this to EL7 for now, CentOS 7 will be needed as well, but not sure if we will need it there right away. @bocekm, would you know?

bstinsonmhk:
This patch looks like an ok start whenever we're ready to ship this package in EL9, feel free to reopen this at that time.

r0x0d:
> This patch looks like an ok start whenever we're ready to ship this package in EL9, feel free to reopen this at that time. 

This patch is not for EL9, it is for EL7. Should I create it in another repository or somewhere else? I get that the label was wrong too, but, could you point out to me the correct place to start out?

bstinsonmhk:
There is nothing further that you need to do to work on that package in RHEL 7. This github repo only applies to EL9 and later.

configs/sst_conversions.yaml
bocekm: We need the worker package in RHEL 7 only. We're working with the Software Production team to have the package copied over to a new unprotected CDN repository with every new version of the package (https://issues.redhat.com/browse/RHELDST-18108). We don't care if the package lands in CentOS Stream, Rocky, Alma, or other distros.

PR #900, opened 2023-06-28
Move XDP workload to networking SST

tohojo:
Update sst_kernel_tps-xdp.yaml to move the workload to the networking sst where it belongs (since I maintain all the mentioned packages anyway).


yselkowitz:
Do you want to rename this file to match?

tohojo:
Ah, lol, didn't even notice the sst name was also in the file name; sorry about that!

Better now? :)

yselkowitz (approved):

PR #904, opened 2023-06-28
Remove Hesiod workload

flo-renaud:
Hesiod has been deprecated since RHEL 7 and was on ACG Level 1 compatibility. It can finally be removed in RHEL10.


jwboyer:
It's still present in CentOS Stream 9 though.  Deleting this file makes it disappearn from the c9s view, which isn't accurate.  I think you just want to remove the `eln` label.

PR #905, opened 2023-06-29
Marking nvme-stas and ktls-utils as unwanted.

rhawalsh:
These packages are going to be delivered via EPEL instead of RHEL in the future.

### DIFF
--- a/configs/sst_cs_apps-containers.yaml
+++ b/configs/sst_cs_apps-containers.yaml
@@ -1,0 +1,47 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Application Streams Container Components
+  description: Components required by RHEL Application Streams container images
+  maintainer: sst_cs_apps
+
+  packages:
+  # common tools used in container scripts and OpenShift tools
+  - bsdtar
+  - bzip2
+  - findutils
+  - gettext-envsubst
+  - git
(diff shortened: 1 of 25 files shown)


configs/sst_storage_io-packages-unwanted.yaml
```diff
@@ -6,8 +6,10 @@ data:
   maintainer: sst_storage_io
 
   unwanted_packages:
+  - ktls-utils
   - libhbaapi
   - libhbalinux
+  - nvme-stas
```
jwboyer, line 12: This is provided in CentOS Stream/RHEL 9 today and isn't going to be removed at this point.  It would be better to create an `eln` specific file and include it there.

rhawalsh: Ah, that makes sense.  I will split the eln and c9s labels and only add these packages to that.  I wasn't sure if existing c9s information was still relevant.  Thank you for confirming.

jwboyer:
Something looks off on this one now.  You have a bunch of additions that look unrelated.  Can you look into why that might be?

rhawalsh:
Ugh, you're right.  I think someone pushed into this PR inadvertently or something.  My local branch doesn't have those changes in it.  I will close this and open a new PR referencing this one to clean things up.

PR #906, opened 2023-06-29
Add RHEL10 and RHEL9 content amendments.

jnovy:
This adds container-tools' related content amendments and preparation for RHEL10.

### DIFF
--- a/configs/sst_container_tools.yaml
+++ b/configs/sst_container_tools.yaml
@@ -5,7 +5,7 @@ data:
   description: >
     Container runtime and related packages maintained by the Container Tools SST
 
-  maintainer: jnovy
+  maintainer: sst_container_tools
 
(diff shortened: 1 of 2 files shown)


configs/sst_container_tools.yaml
```diff
@@ -4,28 +4,38 @@ data:
   name: Container Tools
   description: >
     Container runtime and related packages maintained by the Container Tools SST
-  maintainer: sst_container_tools
+
+  maintainer: jnovy
```
jwboyer, line 8: Why switch this to you instead of leaving it as the team?  The team assignment helps with other pieces.

jnovy: I'm happy to switch it back to sst_container_tools.

PR #908, opened 2023-06-29
Do not ship mariadb-embedded

hhorak:


configs/sst_cs_apps-db-mariadb.yaml
```diff
@@ -14,8 +14,6 @@ data:
   - mariadb-server
   - mariadb-server-galera
   - mariadb-devel
-  - mariadb-embedded-devel
-  - mariadb-embedded
```
jwboyer, line 18: We already ship this in CentOS Stream/RHEL 9 today.  We're not going to remove it, so marking it as unwanted in a file with the `c9s` label is incorrect.  It would be better to create an `eln` specific unwanted file and include these there.

hhorak: Right, thanks for pointing to this, fixed now.

PR #910, opened 2023-06-29
Remove mariadb-connector-c-test, likely not needed to be shipped

hhorak:


hhorak:
This MR is based on https://github.com/minimization/content-resolver-input/pull/908, which needs to be merged before this to avoid merge conflicts.

PR #918, opened 2023-06-30
Update for newly added subpackages in langpacks

pnemade:
Update langpacks workload to include newly added langpacks packages.


juhp:
I think you have to fork to an `-eln` file: most of those are not in c9s/rhel9.

pnemade:
Right forgot to change the labels. Now corrected them.

PR #919, opened 2023-06-30
Add hunspell package and some missing hunspell dictionary packages

pnemade:
The sst_i18n team would like to continue supporting hunspell packages for spellchecking purposes in ELN and upcoming RHEL releases.

### DIFF
--- a/configs/sst_i18n-hunspell-dictionaries-eln.yaml
+++ b/configs/sst_i18n-hunspell-dictionaries-eln.yaml
@@ -5,12 +5,14 @@ data:
   description: A set of hunspell dictionaries
   maintainer: sst_i18n
   labels:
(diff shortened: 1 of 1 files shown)


configs/sst_i18n-hunspell-dictionaries.yaml
```diff
@@ -121,6 +124,7 @@ data:
   - hunspell-tl
   - hunspell-tn
   - hunspell-tpi
+  - hunspell-tr
```
juhp, line 127: This does not seem to be in c9s (rhel9): so I think you need to fork this file.

pnemade: Ok created -eln suffix workload now

PR #936, opened 2023-07-11
Virt SST: Add libblkio and remove libguestfs-inspect-icons

yashkmankad:


yashkmankad:
@bstinsonmhk Hi Brian, this one can be merged. Should be the last of the Virt SST changes before the the Stream 10 import.

PR #937, opened 2023-07-13
Dropping neptune

sandrobonazzola:
Neptune image has been dropped with
https://gitlab.com/CentOS/automotive/sample-images/-/merge_requests/342 There's no need to keep it included within the composes.


barpavel (approved):
LGTM.
I'm not an expert in this repository, but the change makes sense.

PR #947, opened 2023-07-24
crypto: add intel multi buffer libraries

erig0:


yselkowitz:
@erig0 Does this mean it is the intention for this conditional to be enabled in RHEL builds?

https://src.fedoraproject.org/rpms/qatengine/blob/rawhide/f/qatengine.spec#_31

erig0:
I appreciate the quick merge, but these should be arch specific packages, i.e. x86_64 only! My bad. I'll open a follow up PR.

erig0:
> @erig0 Does this mean it is the intention for this conditional to be enabled in RHEL builds?

Yes. That conditional should also be enabled in RHEL.

PR #948, opened 2023-07-24
DRAFT: crypto: intel: packages are arch specific

erig0:
These packages are x86_64 only.

### DIFF
--- a/configs/sst_security_crypto-intel.yaml
+++ b/configs/sst_security_crypto-intel.yaml
@@ -5,10 +5,11 @@ data:
   description: libraries needed for acceleration of crypto and compression on Intel processors
   maintainer: sst_security_crypto
 
-  packages:
-  - intel-ipsec-mb
-  - intel-ipp-crypto-mb
-
   labels:
   - eln
   - c9s
+
+  arch_packages:
+    x86_64:
+      - intel-ipsec-mb
+      - intel-ipp-crypto-mb


configs/sst_security_crypto-intel.yaml
```diff
@@ -5,10 +5,12 @@ data:
...
-  - intel-ipsec-mb
-  - intel-ipp-crypto-mb
-
   labels:
   - eln
   - c9s
+
+  arch_packages:
+    x86_64:
+      - arch-specific-package1
```
yselkowitz, line 14: This line is a typo.

yselkowitz:
@erig0 also there seems to be some confusion about adding these packages to RHEL: https://src.fedoraproject.org/rpms/qatengine/pull-request/3

configs/sst_security_crypto-intel.yaml
erig0: Yikes. copy/paste error. Thanks for catching.

erig0:
> @erig0 also there seems to be some confusion about adding these packages to RHEL: https://src.fedoraproject.org/rpms/qatengine/pull-request/3

Thanks. I'll investigate.

yselkowitz:
The qatengine change which would have used these libraries was rejected.  It seems this can be closed; #954 reverts these additions entirely.

PR #950, opened 2023-07-26
Update subscription-manager & Insights

ptoscano:
See the individual commits for more details.

### DIFF
--- a/configs/sst_insights.yaml
+++ b/configs/sst_insights.yaml
@@ -5,24 +5,46 @@ data:
   description: Insights
   maintainer: sst_insights
 
-  packages: []
+  packages:
+    - yggdrasil
+    - yggdrasil-worker-package-manager
 
   package_placeholders:
-    insights-client:
-      srpm: insights-client
-      description: insights
-      requires:
-      - coreutils
-      - gpg
-      - pciutils
-      - python3-magic
-      - python3-six
-      - python3-requests
-      - systemd
-      - python3-PyYAML
-      - tar
+    - srpm_name: insights-client
+      rpms:
+        - rpm_name: insights-client
+          description: insights-client
+          dependencies:
+            - coreutils
+            - gpg
+            - pciutils
+            - policycoreutils-python-utils
+            - python3-PyYAML
+            - python3-magic
+            - python3-requests
+            - python3-setuptools
+            - python3-six
+            - systemd
+            - tar
+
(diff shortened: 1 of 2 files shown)


yselkowitz:
Please write the placeholders in the new format.

ptoscano:
@yselkowitz anything else that I need to change here?

configs/sst_insights.yaml
```diff
@@ -8,19 +8,56 @@ data:
...
+      - rpm_name: rhc
+        description: rhc
+        dependencies:
+        - policycoreutils-python-utils
+        - yggdrasil
+        - yggdrasil-worker-package-manager
+
+    - srpm_name: rhc-worker-playbook
+      build_dependencies:
+      - rhc
```
yselkowitz, line 39: Placeholders cannot require other placeholders.

configs/sst_insights.yaml
```diff
@@ -8,19 +8,56 @@ data:
...
+      rpms:
+      - rpm_name: rhc-worker-playbook
+        description: rhc-worker-playbook
+        dependencies:
+        - ansible-core
+        - insights-client
```
yselkowitz, line 45: Placeholders cannot require other placeholders.

configs/sst_insights.yaml
```diff
@@ -8,19 +8,56 @@ data:
...
+    - yggdrasil-worker-package-manager
+
+    ppc64le:
+    - yggdrasil-worker-package-manager
+
+    s390x:
+    - yggdrasil-worker-package-manager
+
+    x86_64:
+    - yggdrasil-worker-package-manager
```
yselkowitz, line 60: `arch_packages` is for when a package is only for some architectures.  (It is also not needed just for excluding i686.)  What are you excluding here?

yselkowitz (changes requested):

configs/sst_insights.yaml
ptoscano: ah ok; from what I remember in the past (during the el9 work), `packages` was for noarch packages and `arch_packages` was for arch-specific ones

in this case, I can also fold the `arch_packages` in `sst_subscription_manager-subscription-manager.yaml` to `packages`

ptoscano:
> Placeholders cannot require other placeholders.

maybe `test_config_files.py` could check that? ;)

PR #951, opened 2023-07-31
Make it easier to detect that GHC was included by accident

fweimer-rh:
GHC uses versioned packages (ghc9.8), so the previous unwanted package entry did not match it.  List a few common GHC packages as well: alex (lexer generator), happy (parser generator), and ghc-bytes, ghc-bytes (libraries fairly central to binary data processing).

### DIFF
--- a/configs/sst_program-unwanted.yaml
+++ b/configs/sst_program-unwanted.yaml
@@ -50,6 +50,7 @@ data:
     - koji-utils
     - koji-vm
     - koji-web
+    - pandoc
     - python3-koji
     - python3-koji-cli-plugins
     - python3-koji-hub


yselkowitz:
Whichever, but separate may make more sense.

configs/sst_program-unwanted.yaml
```diff
@@ -50,6 +50,7 @@ data:
     - koji-utils
     - koji-vm
     - koji-web
+    - pandoc
```
jwboyer, line 53: This doesn't look correct.  I understand pandoc in Fedora is built so that it requires GHC, but in RHEL we build it differently so it doesn't get GHC pulled in.

@yselkowitz is there a way to stub out pandoc with a placeholder package to reflect how we build it in RHEL instead of marking it unwanted entirely?

jwboyer: Nevermind, I am mis-remembering what we did in RHEL 9.  This is correct.

PR #952, opened 2023-08-03
Gate awscli2 out of s390x for the Meta ELN Extras config

davide125:
This package has an ExcludeArch on s390x so we need to gate it out to unbreak the workload.

### DIFF
--- a/configs/eln_extras_meta.yaml
+++ b/configs/eln_extras_meta.yaml
@@ -96,11 +96,6 @@ data:
       - awscli2
       - sedutil
       - sysbench
-    i686:
-      - awscli2
-      - msr-tools
-      - sedutil
-      - sysbench
     ppc64le:
       - awscli2
       - sedutil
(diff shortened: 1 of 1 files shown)


configs/eln_extras_meta.yaml
```diff
@@ -94,15 +93,19 @@ data:
 
   arch_packages:
     aarch64:
+      - awscli2
       - sedutil
       - sysbench
     i686:
+      - awscli2
```
yselkowitz, line 100: why is i686 here at all?  (otherwise looks good though)

davide125: Uh, I actually don't know, I think I copied the structure from another config. We definitely don't need i686 for what we're using this config at Meta. Should I drop it?

yselkowitz: Yes, drop the entire i686 field here (probably makes more sense as a separate commit as part of this PR).

yselkowitz (approved):

PR #953, opened 2023-08-03
Remove ESC package from base RHEL

Jakuje:
The ESC is a tool for RHCS smart cards enrollment and it should be moved to the RHCS layered product. It has very limited use in base RHEL.

### DIFF
--- a/configs/ssg_idm-smart-cards-unwanted.yaml
+++ b/configs/ssg_idm-smart-cards-unwanted.yaml
@@ -1,12 +1,12 @@
-document: feedback-pipeline-workload
+document: feedback-pipeline-unwanted
 version: 1
 data:
-  name: IdM Smart Cards (RHCS specific)
+  name: IdM Smart Cards unwanted (RHCS specific)
   description: Specific packages for client-side smart card enrollment on RHEL.
   maintainer: ssg_idm
 
-  packages:
-  - esc
+  unwanted_packages:
+    - esc
 
   labels:
-  - c9s
+    - eln


jwboyer:
This doesn't seem correct.  `esc` is still shipped in RHEL 9 and can't be moved at this point.  If you want to make that transition for future RHEL releases, then you should simply delete the `eln` label from the file and then create a new unwanted file with the `eln` label that lists `esc`

Jakuje:
Thank you for the comment! It was some time since I was changing something in the resolver so sorry for not getting it on the first try. Hope the second attempt with separate "unwanted" file will do.

configs/ssg_idm-smart-cards-unwanted.yaml
```diff
@@ -0,0 +1,12 @@
+document: feedback-pipeline-unwanted
+version: 1
+data:
+  name: IdM Smart Cards unwanted (RHCS specific)
+  description: Specific packages for client-side smart card enrollment on RHEL.
+  maintainer: ssg_idm
+
+  packages:
```
yselkowitz, line 8:
```suggestion
  unwanted_packages:
```

yselkowitz (changes requested):

configs/ssg_idm-smart-cards-unwanted.yaml
Jakuje: obviously! Fixed now.

yselkowitz (approved):

jwboyer:
Thanks for the quick fixes @Jakuje and the review @yselkowitz

PR #954, opened 2023-08-03
Revert intel MB crypto, fix vfxplatform

yselkowitz:
- Revert "crypto: add intel multi buffer libraries"
- Fix ELN VFX Platform dependencies

/cc @erig0 @jwakely


erig0:
We can revert the crypto changes for now. However, we may still attempt to bring ipsec-mb and ipp-crypto-mb to RHEL, but they need more investigation first.

PR #955, opened 2023-08-03
cloud experience: add efs-utils

major:
This package adds integrations for Elastic File System (EFS) on AWS which allows users to mount and manage EFS much more efficiently.

### DIFF
--- a/configs/sst_cloud_experience.yml
+++ b/configs/sst_cloud_experience.yml
@@ -1,0 +1,14 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Cloud Experience Integrations for RHEL
+  description: Software that makes RHEL more integrated with various clouds
+  maintainer: sst_cloud_experience
+
+  packages:
+    - efs-utils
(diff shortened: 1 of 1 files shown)


yselkowitz:
This appears to depend on python3-botocore, which was in RHEL 8 HA but not in 9.  Will your SST be maintaining that as well?

major:
Good catch, @yselkowitz. That slipped by me as I was testing on a system with EPEL enabled. 🤦‍♂️

Let me sync with the team and see how we want to approach this one.

yselkowitz:
Could you please mark this as draft in the meantime?

major:
Done. Thank you.

major:
@yselkowitz Updated and python-botocore is added in there.

major:
@yselkowitz My goal would be to get it into c9s/RHEL9. Sorry, this is my first one of these and I'm still reading through all of the docs.

yselkowitz (approved):
Thanks for clarifying.

PR #956, opened 2023-08-08
Fedora no longer has modular repos.

tdawson:
Remove them from our configs.


sgallagher:
Aww, I'm going to miss "Rawhdular"

PR #963, opened 2023-08-24
Cleanup network workloads

tyll:
Various cleanups for the networking workloads

/cc @smwfaye


yselkowitz:
@bstinsonmhk this workload is not showing up at all in Content Resolver, causing several packages to be erroneously dropped from ELN.  There is no obvious reason as to why though.

asamalik:
Heh it doesn't see it because it's `.yml` and Content Resolver only sees `.yaml`

PR #970, opened 2023-09-01
Add post-quantum crypto: liboqs, oqs-provider

neverpanic:
Since there is industry demand for post-quantum cryptography, add the recently packaged liboqs and oqs-provider packages as requested for eln.

### DIFF
--- a/configs/sst_security_crypto-pqc.yaml
+++ b/configs/sst_security_crypto-pqc.yaml
@@ -1,0 +1,12 @@
+---
+document: feedback-pipeline-workload
+version: 1
+data:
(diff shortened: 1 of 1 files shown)


yselkowitz:
@neverpanic CR is not finding the oqs-provider package, is that going to be added to Fedora or will it be RHEL only (and therefore need a placeholder instead)?

neverpanic:
@yselkowitz It should already be part of Fedora. It seems I typo'd and need to delete the dash: https://src.fedoraproject.org/rpms/oqsprovider

neverpanic:
Fix in #971.

PR #971, opened 2023-09-03
post-quantum crypto: Fix package name typo

neverpanic:
The OpenSSL OQS Provider is packaged in oqsprovider, not oqs-provider.

### DIFF
--- a/configs/sst_security_crypto-pqc.yaml
+++ b/configs/sst_security_crypto-pqc.yaml
@@ -7,6 +7,6 @@ data:
   maintainer: sst_security_crypto
   packages:
     - liboqs
-    - oqs-provider
+    - oqsprovider
   labels:
     - eln


yselkowitz (approved):

neverpanic:
@jwboyer @yselkowitz I see oqsprovider and liboqs in `buildroot` now according to [\[1\]][1] and [\[2\]][2], but we want the package to be available to users in BaseOS. Is that something I can do at this point, and if so, how?

[1]: https://tiny.distro.builders/view-rpm--view-eln--oqsprovider.html
[2]: https://tiny.distro.builders/view-rpm--view-eln--liboqs.html

yselkowitz:
> @jwboyer @yselkowitz I see oqsprovider and liboqs in `buildroot` now according to [[1]](https://tiny.distro.builders/view-rpm--view-eln--oqsprovider.html) and [[2]](https://tiny.distro.builders/view-rpm--view-eln--liboqs.html), but we want the package to be available to users in BaseOS. Is that something I can do at this point, and if so, how?

@sgallagher is this related to the recent comps changes?

sgallagher:
Doesn't look [like it](https://pagure.io/pungi-fedora/c/c8e199e00c972a489c8ab246b53c42083571f807?branch=eln). It wasn't previously in [`prepopulate.json`](https://pagure.io/pungi-fedora/blob/eln/f/prepopulate.json) which means we don't know which repo to put it in. That means it defaults to ending up in the buildroot (so at least it's guaranteed to be available to build against).

We're in the middle of figuring out how to design an extension to Content Resolver to allow individuals to propose the correct repo. In the meantime, the non-scalable option is to ask @jwboyer to do it for you.

PR #972, opened 2023-09-04
Add CMake to the VFX Platform

Conan-Kudo:
This is a required tool for building VFX software and needs to be tracked alongside other components required for the VFX Platform.

### DIFF
--- a/configs/sst_pt_libraries-vfxplatform-c9s.yaml
+++ b/configs/sst_pt_libraries-vfxplatform-c9s.yaml
@@ -7,6 +7,12 @@ data:
 
   packages:
   # The minimum development environment is this:
+  - cmake
+  - cmake-filesystem
+  - cmake-gui
+  - cmake-data
+  - cmake-doc
+  - cmake-rpm-macros
   - gcc
   - gcc-c++
   - binutils
--- a/configs/sst_pt_libraries-vfxplatform-eln.yaml
+++ b/configs/sst_pt_libraries-vfxplatform-eln.yaml
@@ -7,6 +7,12 @@ data:
 
   packages:
   # The minimum development environment is this:
+  - cmake
+  - cmake-filesystem
+  - cmake-gui
+  - cmake-data
+  - cmake-doc
+  - cmake-rpm-macros
   - gcc
   - gcc-c++
   - binutils


bstinsonmhk:
@codonell Do you want to take a look at this? 

I'm inclined to hold this PR open while we negotiate maintainership for cmake. We're looking at the initial maintainership assignment rounds soon, and I want this to show up on one that needs some discussion.

codonell:
Why do we want to add CMake to the workload?

Components which are part of VFX are listed in the standard: https://vfxplatform.com/

It is an implementation detail that those components need CMake.

The content resolver itself knows that the components in the list need CMake, and that should be sufficient.

As such I'd say it is an over-specification to add CMake to the workload itself, like it would ban over-specification to add anything else that is a secondary dependency of anything in the list of requirements.

Conan-Kudo:
> Components which are part of VFX are listed in the standard: https://vfxplatform.com/
> It is an implementation detail that those components need CMake.

I will bring up with the VFX folks to include a CMake minimum version too, since that defines the usability of other components. It _is_ effectively required and is tracked as part of the VFX Platform developer experience.

> As such I'd say it is an over-specification to add CMake to the workload itself, like it would be an over-specification to add anything else that is a secondary dependency of anything in the list of requirements.

I would agree if it wasn't that people were expected to use CMake to build stuff using it and that it was what the VFX community does everything with. Not to mention we had a scare already where CMake was declared unwanted (4d3c94b4e852d949ef329cdaffa0d6cd009d23d9). CMake is essentially a critical dependency and _must_ be accounted for.

codonell:
> > Components which are part of VFX are listed in the standard: https://vfxplatform.com/
> > It is an implementation detail that those components need CMake.
> 
> I will bring up with the VFX folks to include a CMake minimum version too, since that defines the usability of other components. It _is_ effectively required and is tracked as part of the VFX Platform developer experience.

Thank you. I know it's a bit more work, but it gets everyone on the same page, including CMake features required.

It would be my preference that the VFX standard explicitly call out CMake and the version required (or features).

> > As such I'd say it is an over-specification to add CMake to the workload itself, like it would be an over-specification to add anything else that is a secondary dependency of anything in the list of requirements.
> 
> I would agree if it wasn't that people were expected to use CMake to build stuff using it and that it was what the VFX community does everything with. Not to mention we had a scare already where CMake was declared unwanted ([4d3c94b](https://github.com/minimization/content-resolver-input/commit/4d3c94b4e852d949ef329cdaffa0d6cd009d23d9)). CMake is essentially a critical dependency and _must_ be accounted for.

I'm less concerned because all of the dependencies still exist and you can't just remove CMake.

Example:
https://tiny.distro.builders/view-rpm--view-eln--cmake.html

All of us using the content resolver data should try to find an owner for the package and there should be package ownership discussions happening around that data.

The fact that we're having a conversation right now about workloads is IMO the greatest benefit of the content resolver data set.

sgallagher:
Can we close this MR? CMake is absolutely not being removed from RHEL and the maintainership question is or will be resolved elsewhere.

For clarity, the "unwanted" status mentioned above was a misunderstanding that should not have landed in the CR input and has been [reverted](https://github.com/minimization/content-resolver-input/commit/2a24c094cd691172a2b461eeeacc939fdd572cc8).

Conan-Kudo:
I still believe it should be tracked as part of the VFX workload. It's important to the VFX community to have it maintained and regularly updated, and there is no other obvious way to signal that it's being tracked for that purpose.

yselkowitz:
This team isn't going to own cmake in RHEL 10, and cmake itself is still not listed on the VFX Reference Platform homepage.

PR #976, opened 2023-09-11
Include pkcs11-provider in ELN

neverpanic:
OpenSSL 3.x deprecated the ENGINE API, so add pkcs11-provider for modern OpenSSL PKCS#11 support.

### DIFF
--- a/configs/sst_security_crypto-pkcs11.yaml
+++ b/configs/sst_security_crypto-pkcs11.yaml
@@ -1,0 +1,11 @@
+---
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: PKCS#11 support
+  description: Components providing access to PKCS#11 cryptographic tokens.
+  maintainer: sst_security_crypto
+  packages:
+    - pkcs11-provider
+  labels:
+    - eln


yselkowitz:
@neverpanic in order to replace openssl-pkcs11 with pkcs11-provider in ELN, the following packages need to be modified in rawhide:

- [ ] bind-dyndb-ldap
- [ ] freeipa
- [ ] libssh
- [ ] openssl

Also for Fedora, the following additional packages are affected:

- [ ] rng-tools

These changes may need to be conditionalized on .e.g %if 0%{?fedora} >= 40 || 0%{?rhel} >= 10 to avoid breakage to stable branches.

neverpanic:
@yselkowitz I'm aware of that. We will take care of that in a follow-up ticket and file appropriate tickets for the maintainers of these packages soon. We need the alternative (pkcs11-provider) to be available so that these applications can switch, though.

yselkowitz:
CR input does not block dependencies.  Therefore, while we do want this change and much appreciate your proactiveness, it does not block migration work.  (If this were a leaf, then you would be correct that it would not be included without first being listed.)

neverpanic:
We're tracking the follow-up for those components in CRYPTO-11788.

PR #982, opened 2023-09-20
Mark JMC and it's related packages as unwanted

Josh-Matsuoka:
As part of the process for removing JMC going forward this PR marks jmc, jmc-core, and owasp-java-encoder (a dependency of jmc-core that was still floating around) as unwanted to prevent them from being pulled in going forward.


yselkowitz:
https://github.com/minimization/content-resolver-input/blob/main/configs/sst_java-jmc-eln.yaml needs to be removed as well.

Josh-Matsuoka:
Removed the eln configuration for jmc.

yselkowitz (approved):
I'm not sure how much it's actually needed to mark these as unwanted given that they have all been retired in Fedora, but I guess it at least documents the fact that they're not missing by accident. :man_shrugging:

PR #986, opened 2023-09-29
edge: Add bootupd

cgwalters:
So right now the "coreos" workload is disabled in ELN: https://github.com/minimization/content-resolver-input/blob/2331fc87220bebfe32dc5db5ec038c3948197e46/configs/sst_coreos.yaml#L35

In practice it seems like everything ended up here under edge.

Of tier 0, just missing bootupd, but we want to ship that in Edge.

### DIFF
--- a/configs/sst_edge.yaml
+++ b/configs/sst_edge.yaml
@@ -35,6 +35,7 @@ data:
     - fdo-rendezvous-server
     - greenboot
     - greenboot-default-health-checks
+    - bootupd
     x86_64:
     - clevis-pin-tpm2
     - fdo-client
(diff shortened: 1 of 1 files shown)


cgwalters:
xref

- https://gitlab.com/CentOS/cloud/sagano/-/merge_requests/32
- https://gitlab.com/CentOS/cloud/sagano/-/merge_requests/33

cgwalters:
Ah right this needs to be in `arch_packages`, while here I've also updated for the fact we build on ppc64le now.

cgwalters:
(Honestly maybe it's simpler if we build, but just error out on s390x for now...may look at that upstream)

cgwalters:
Alternative in https://github.com/minimization/content-resolver-input/pull/987

bstinsonmhk:
Is 987 a superseding change? It's OK if we want to keep this in multiple workloads.

cgwalters:
> Is 987 a superseding change?

Yes.

> It's OK if we want to keep this in multiple workloads.

Yeah, we already have that situation today.

PR #987, opened 2023-09-29
sst_coreos: Enable for ELN

cgwalters:
To gain bootupd mainly, but we probably also want
the console-login-helper stuff and `ssh-key-dir` is currently an implicit Ignition dependency etc.


cgwalters:
Alternative to https://github.com/minimization/content-resolver-input/pull/986

cgwalters:
How long does this take to sync?  Looks like https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/ didn't pick this up.

cgwalters:
I see it in https://tiny.distro.builders/view-all-binary-package-name-list--view-eln.txt ...yet today's 2023-10-04 compose doesn't have it either?

cgwalters:
Sorry to nag but anything blocking this showing up in ELN still?  I don't see it in the 2023-10-06 compose https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/

sgallagher:
There's a disconnect between Content Resolver and the compose that we still haven't resolved. We don't have an automatic way to tell the compose which repository each package is supposed to end up in, so we maintain it manually in https://pagure.io/pungi-fedora/blob/eln/f/prepopulate.json

If not listed there, they end up in the buildroot only. @bstinsonmhk has been working on a proposal for how to automate this properly, but for the moment we still need to update prepopulate.json.

sgallagher:
I've submitted https://pagure.io/pungi-fedora/pull-request/1203 to add bootupd to AppStream, but I also want to note that `rust-bootupd` is currently FTBFS on Fedora ELN: https://koji.fedoraproject.org/koji/taskinfo?taskID=107122332

yselkowitz:
The build failure is due to premature unconditional use of the latest Fedora rust macros.  I'll try to work on this when I come back.

cgwalters:
Looks like this is working now right?  https://koji.fedoraproject.org/koji/buildinfo?buildID=2302687

sgallagher:
I've merged the pungi-fedora change to get this properly into AppStream for the next compose. Composes happen four times a day, so you should see this appear in a few hours.

PR #988, opened 2023-10-02
Replace Cheese with Snapshot as the default webcam application in RHEL

tpopela:
Fedora Workstation will make a switch in F40, see
https://pagure.io/fedora-workstation/issue/166 and RH internal https://issues.redhat.com/browse/DESKTOP-739


kalev:
+1 from me

PR #991, opened 2023-10-04
Add Tomcat and mod_proxy_cluster, both for c9s and c10s.

notroj:
For review by @csutherl


configs/sst_middleware-tomcat.yaml
```diff
@@ -0,0 +1,22 @@
+document: feedback-pipeline-unwanted
```
bstinsonmhk, line 1: It looks like you're making a workload so changing the document type to `feedback-pipeline-workload` should fix the failing CI checks.

notroj: Ah, I was staring at the CI failure in confusion, thanks a lot @bstinsonmhk . Repushed like that.

configs/sst_middleware-tomcat.yaml
```diff
@@ -0,0 +1,22 @@
...
+  packages:
+  - tomcat
+  - tomcat-admin-webapps
+  - tomcat-docs-webapp
+  - tomcat-el-3.0-api
+  - tomcat-jsp-2.3-api
+  - tomcat-lib
+  - tomcat-servlet-4.0-api
+  - tomcat-webapps
+  - mod_proxy_cluster
```
yselkowitz, line 17: mod_proxy_cluster is currently a placeholder in sst_cs_infra_services-httpd; as it was just added to rawhide, it should be a real package now, question is in which workload/SST it belongs.

yselkowitz:
Also, does this mean that the RHEL-specific references to `pki-servlet-engine` in dogtag-pki and tomcatjss spec files will be removed?

notroj:
> Also, does this mean that the RHEL-specific references to `pki-servlet-engine` in dogtag-pki and tomcatjss spec files will be removed?

Yes they should be removed or versioned to be RHEL < 10 specific

notroj:
At least I assume so, will chat with @csutherl about it

configs/sst_middleware-tomcat.yaml
notroj: Thanks. Fixed in  f10aa2b - `sst_middleware` owns this package in RHEL 8+,

PR #993, opened 2023-10-05
Mark python3-py as unwanted in RHEL 10+

befeleme:


configs/sst_cs_apps-unwanted-python3-eln.yaml
```diff
@@ -16,6 +16,8 @@ data:
   # tomllib is in the Python standard library (PEP 680), use that one instead
   - python3-toml
   - python3-tomli
+  # python3-py was needed by pytest which bundled the necessary parts
```
hroncok, line 19: What about this?

```suggestion
  # python3-py is in maintenance mode upstream, pretty much dead
  # it used to be in RHEL as a dependency of pytest which bundled the necessary parts
```

configs/sst_cs_apps-unwanted-python3-eln.yaml
```diff
@@ -27,6 +29,7 @@ data:
   unwanted_source_packages:
   - pytz
   - python-chardet
+  - python-py
   - python-toml
   - python-tomli
```
hroncok, line 34: Perhaps keep the same order here?

```suggestion
  - python-toml
  - python-tomli
  - python-py
```

hroncok (commented):
Ack for the actual content change. 2 nitpicks.

befeleme:
Thank you. I accepted both changes.

hroncok (approved):
Thanks!

PR #997, opened 2023-10-06
sst_coreos: bootupd is x86_64/aarch64 only

yselkowitz:
Based on the upstream documentation, bootupd is not supported on ppc64le, and is not built therefor in either Fedora or c9s.

/cc @cgwalters

### DIFF
--- a/configs/sst_coreos.yaml
+++ b/configs/sst_coreos.yaml
@@ -31,8 +31,6 @@ data:
     - bootupd
     aarch64:
     - bootupd
-    ppc64le:
-    - bootupd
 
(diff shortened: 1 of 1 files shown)


cgwalters:
Upstream docs are out of date, we support ppc64le now; kicked off https://src.fedoraproject.org/rpms/rust-bootupd/c/7a9c901f8bd25fc80f5cc4e10d2ea4f400492c82?branch=rawhide

cgwalters:
Perhaps at this point I should just enable all arches but make it no-op on s390x for now...

PR #999, opened 2023-10-09
Drop Evolution from RHEL 10

tpopela:
* Evolution was appreciated by customers for its excellent Microsoft Exchange support, but Thunderbird's Exchange support was subpar, but they're now investing in resolving that
 * Microsoft recently announced that they will only allow Graph API for accessing email on Office365 and we don't support it in Evolution as of now (this will affect older RHEL releases as well)
 * We cannot justify to maintain 2 separate desktop email clients in RHEL
 * Evolution Data Server will stay as its the integral part of the GNOME
 * Evolution is and will be available on Flathub
 * Removing Evolution will allow us to remove WebKitGTK from RHEL 10

See Red Hat internal https://issues.redhat.com/browse/DESKTOP-797 for more information.


Conan-Kudo:
> Evolution was appreciated by customers for its excellent Microsoft Exchange support, but Thunderbird's Exchange support was subpar, but they're now investing in resolving that.

This is news to me. Any information on this front?

configs/sst_desktop_applications-evolution-data-server.yaml
```diff
@@ -0,0 +1,15 @@
...
+data:
+  name: Evolution Data Server
+  description: Unified backend for programs that work with contacts, tasks, and calendar information
+  maintainer: sst_desktop_applications
+
+  packages:
+  - evolution-data-server
+  - evolution-data-server-langpacks
+  - evolution-ews
+  - evolution-ews-langpacks
```
yselkowitz, line 12: @tpopela please note that evolution-ews includes both e-d-s and evolution plugins, and its continued presence means that evolution itself is still being pulled into ELN.  There does not seem to be a built-in way to build only the e-d-s components.  Furthermore, given your explanation in this PR wrt Exchange support, shouldn't this too be retired, and only e-d-s itself remain?

tpopela:
We're aware of this and working on it.

Dne st 18. 10. 2023 23:34 uživatel Yaakov Selkowitz <
***@***.***> napsal:

PR #1000, opened 2023-10-09
Drop WebKitGTK from RHEL 10

tpopela:
The only supported way (albeit not embeddable) in RHEL to render HTML content is Firefox. See Red Hat internal
https://issues.redhat.com/browse/DESKTOP-670 for more information.


Conan-Kudo:
😢

PR #1002, opened 2023-10-16
Mark Inkscape as unwanted

tpopela:
Removing the workload for ELN wasn't enough as some packages (such as djvulibre of python-matplotlib) are pulling it through build requires.


yselkowitz:
No objections to this, but for the record, inkscape is currently being pulled in as an indirect dependency of other unwanted build dependencies.  There is no work needed to specifically remove inkscape from wanted packages.

tpopela:
That's true, but I wanted to be explicit, as before this change if someone opened the CR page for Inkscape they might get the false notion that it might show up in RHEL 10.

PR #1006, opened 2023-10-21
bootc: New sst

cgwalters:
The subsystem is new and the package is
just built in fedora, xref https://bugzilla.redhat.com/show_bug.cgi?id=2243345

Signed-off-by: Colin Walters <walters@verbum.org>

---


cgwalters:
Attempted followup here https://pagure.io/pungi-fedora/pull-request/1217

PR #1007, opened 2023-10-23
Add sysprof to ELN/RHEL 10

tpopela:
Finally introduce a tool that can help with profiling across the whole system.


yselkowitz (changes requested):
libunwind is still marked unwanted, so that should be fixed too.

davide125:
Note that for the time being ELN disables frame pointers, which is a deviation from Fedora (https://src.fedoraproject.org/rpms/redhat-rpm-config/c/9e10bb665ec2d3cc4b1ea613a7fe8e99e467e64e?branch=rawhide). This makes sysprof significantly less useful in practice as you can't get good stacktraces out of it unless you recompile a large part of the system. If you plan to ship sysprof in RHEL 10 I would recommending reverting that change and enabling frame pointers there as well. cc @rwmjones @chergert who can speak to the specifics of this as well.

chergert:
While I would *love* for RHEL to enable frame-pointers, that is completely outside my control.

That said, if we could at least enable our engineers to see something that is good. When we don't have frame-pointers, sometimes we can "invert" the stack trace to merge at the leaves to figure some things out, even if less precise.

Additionally, we can give some special builds to customers which contain enough to deduce things even if the stack traces are rather horrible.

It's an imprecise art, but it's better than nothing.

configs/sst_desktop-sysprof.yaml
```diff
@@ -0,0 +1,12 @@
...
+data:
+  name: Sysprof
+  description: System-wide profiling tool
+  maintainer: sst_desktop
+
+  packages:
+  - sysprof
+
+  labels:
+  - eln
```
Conan-Kudo, line 12: Shouldn't this also be marked for `c10s` too?

Conan-Kudo (changes requested):

configs/sst_desktop-sysprof.yaml
tpopela: c10s is still eln, no? I not, then I'm not aware as there wasn't any RH internal or Fedora announcement.

Conan-Kudo: It should be, but then I started seeing `c10s` labels show up on some of the yamls. I don't know what's going on...

yselkowitz: Some people are just jumping the gun, that's all.  c10s is not yet split from eln, and when that split does occur, any configs with eln will get a c10s label added.  No harm in adding it now, but it's not required yet.

yselkowitz (approved):

PR #1010, opened 2023-10-25
Remove xsane from c10s

zdohnal:
Do not ship xsane until it is ported to GTK3.


yselkowitz:
https://github.com/minimization/content-resolver-input/blob/main/configs/sst_cs_infra_services-scanning-stack.yaml needs to be split, dropping xsane from the eln version.

zdohnal:
Thank you for the review! I've incorporated the feedback - does it look good now?

PR #1012, opened 2023-10-29
sst_cs_plumbers-editors: emacs-nox -> emacs-nw

Conan-Kudo:
The emacs-nox package was renamed to emacs-nw, presumably to match the actual name of the option and to drop the tight association with X11 when this also applies to non-X11 windowing environments (such as Wayland).


bstinsonmhk:
@lnykryn any thoughts about this?

jacekmigacz:
(on @lnykryn request and as emacs maintainer for RHEL)
Since c10s and rawhide emacs has already %package nw is looks perfectly fine to me.

PR #1014, opened 2023-10-31
Mark motif as an unwanted package

nielsdg:
Motif has been deprecated in RHEL 9 (see [1] and [2]) and technically even already in the RHEL 8 lifecycle (see [3]), and we're now removing this in RHEL 10.

See also Red Hat internal tracking ticket
https://issues.redhat.com/browse/GPU-621

[1]: https://access.redhat.com/solutions/6113101
[2]: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/9.0_release_notes/deprecated_functionality#JIRA-RHELPLAN-98983
[3]: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/8.5_release_notes/deprecated_functionality#JIRA-RHELPLAN-98983

### DIFF
--- a/configs/sst_gpu-X11-enviroment-c9s.yaml
+++ b/configs/sst_gpu-X11-enviroment-c9s.yaml
@@ -12,5 +12,4 @@ data:
     - xrestop
 
   labels:
-    - eln
     - c9s
--- a/configs/sst_gpu-X11-enviroment-eln.yaml
+++ b/configs/sst_gpu-X11-enviroment-eln.yaml
@@ -1,0 +1,14 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Traditional X11 environment
+  description: Libraries and sample applications historically provided with an X11 installation
+  maintainer: sst_gpu
+
+  packages:
+    - libXaw
+    - mesa-libGLU
+    - xrestop
+
+  labels:
+    - eln
--- a/configs/sst_gpu-unwanted-c9s.yaml
+++ b/configs/sst_gpu-unwanted-c9s.yaml
@@ -1,0 +1,48 @@
+document: feedback-pipeline-unwanted
+version: 1
+data:
+  name: GPU - Blocked packages
+  description: GPU - Packages which should not be in RHEL 9
+  maintainer: sst_gpu
+  unwanted_packages:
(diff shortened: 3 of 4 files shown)


yselkowitz:
motif is currently being pulled into ELN only by mesa-libGLw in https://github.com/minimization/content-resolver-input/blob/main/configs/sst_gpu-X11-enviroment.yaml .  That config will need to be split, removing mesa-libGLw from the eln config.

nielsdg:
Yep, that's correct. We've been discussing this internally, and are probalby going to remove mesa-libGLw entirely. I wasn't sure if it made sense to do that in a separate PR though

yselkowitz:
Please include it here, because it doesn't make sense to mark motif as unwanted while mesa-libGLw is still listed as wanted.

nielsdg:
Makes sense, did that in a followup commit which I just added.

configs/sst_gpu-X11-enviroment.yaml
```diff
@@ -7,7 +7,6 @@ data:
 
   packages:
     - libXaw
-    - mesa-libGLw
```
yselkowitz, line 10: This config is for both c9s and eln, and we can only remove this from ELN at this point.  Therefore, as previously mentioned, this config will need to be split; iow separate files for c9s and eln, with this package listed only in the c9s config.

yselkowitz (changes requested):

configs/sst_gpu-X11-enviroment.yaml
nielsdg: ah, sorry about that. Should be fixed now!

configs/sst_gpu-unwanted.yaml
```diff
@@ -31,6 +31,8 @@ data:
   - imake
   # Only used by optional plugin of hwloc, removed for RHEL.
   - libXNVCtrl
+  # Deprecated since RHEL 9 (technically even 8.4), removed in RHEL 10
+  - motif
```
bstinsonmhk, line 35: Since motif is still in RHEL 9 (even though it's deprecated), you'll want to split the unwanted config too

nielsdg: oh, good catch, I hadn't thought of that! Should be fixed now :-)

PR #1015, opened 2023-11-01
Add perl-Digest-SHA1 to unwanted Perl packages

michal-josef-spacek:
We have perl-Digest-SHA, which owns SHA1 algorithm


yselkowitz:
Please note that perl-Digest-SHA1 is a runtime dependency of perl-Net-SNMP (a transitive dependency of mrtg) and a development dependency of spamassassin.  These dependents will need to be fixed in order for perl-Digest-SHA1 to be removed from ELN.

michal-josef-spacek:
@yselkowitz I am working on the fix in perl-Net-SNMP actually

yselkowitz (changes requested):
This config currently covers both c9s and eln.  If SHA1 is in RHEL9 then it cannot be removed, and split configs will be needed for each of c9s and eln.

michal-josef-spacek:
@yselkowitz Ah good point. I created copy. And remove perl-Socket6 and perl-IO-Socket-INET6 from c9s (previous commit).

yselkowitz (approved):

PR #1016, opened 2023-11-01
Add exfatprogs

preichl:
exfatprogs was newly added to RHEL-9, no idea why it is not supposed to be part of RHEL-10


yselkowitz:
exfatprogs was being pulled in as a dependency of udisks2, but that dependency [was softened recently](https://src.fedoraproject.org/rpms/udisks2/c/92fceb0c5dabd8f7845a1648d3e5ff3140e16edc?branch=rawhide), which caused it to fall off of Content Resolver.

FWIW, udftools is in the same boat.  What is the plan for that in RHEL 10?

preichl:
I am sorry I know very little about this effort, I just want to make sure that we deliver exfatprogs in rhel-10, no matter if there are any dependencies or not. Users/customers simply want it.

preichl:
@jwboyer Hello, can you merge the PR, please?

PR #1019, opened 2023-11-02
Move Thunderbird and Firefox from ELN to ELN Extras

tpopela:
We're using different packaging in RHEL (and in RHEL 10 those will be shipped as Flatpaks only) than in Fedora so the dependencies are not the same and the ELN is being "polluted" by packages that won't be needed in RHEL at all. On the other hand we're still interested in knowing how Thundebird and Firefox comply with all the ELN specific changes (extra hardening flags, ..) so ELN Extras is better place for these.

### DIFF
--- a/configs/sst_desktop_applications-firefox.yaml
+++ b/configs/sst_desktop_applications-firefox.yaml
@@ -9,5 +9,10 @@ data:
   - firefox
 
   labels:
-  - eln
-  - c9s
+  # This package is only in ELN Extras because in RHEL we're using a different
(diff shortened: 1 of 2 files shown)


tpopela:
@sgallagher once this is merged, then you can remove Firefox and Thunderbird from the distrobaker config to not sync them into c10s.

sgallagher (approved):

sgallagher:
Related: https://pagure.io/pungi-fedora/pull-request/1219

PR #1020, opened 2023-11-03
bootc: Add composefs to c10s/eln

cgwalters:
ostree and podman are going to centralize on this as a key underlying technology.

### DIFF
--- a/configs/sst_bootc-eln.yaml
+++ b/configs/sst_bootc-eln.yaml
@@ -8,6 +8,7 @@ data:
 
   packages:
   - bootc
+  - composefs
 
   labels:
   - c10s


giuseppe (approved):
LGTM

cgwalters:
OK so a month later...composefs still isn't in ELN, I guess because it may be the prepopulate.json hasn't been updated?

But there's a lot of other PRs here that aren't updating the prepopulate?

cgwalters:
In the meantime, I tried to bypass this by just building composefs in copr for ELN.  However *that* also breaks because apparently copr can't find `go-md2man` because it isn't in ELN?  Or I guess really it's only in a special "buildroot only" set of things?  The pain level here is really really high...

cgwalters:
Ahhh wow I see now, podman was sufficiently burned by this that it just vendors its own copy of go-md2man and builds it at build time.

Wait ok but netavark also uses go-md2man and the [copr build is somehow finding it](https://download.copr.fedorainfracloud.org/results/rhcontainerbot/podman-next/fedora-eln-x86_64/06726861-netavark/builder-live.log.gz)...perhaps there is some special flag to enable the special "ELN build only repos"?

cgwalters:
Update: apparently the secret sauce here is to click the "edit" button on each individual arch repo and copy/paste the URL to the build repo.

(Why copr doesn't have a handy high level opt-in "enable distribution buildroot" or something, dunno...or even better some variant of that that can be opted-into declaratively)

PR #1025, opened 2023-11-21
Remove Initial Setup and anaconda-user-help packages from RHEL-10

jkonecny12:
See reasoning in commits.


poncovka (approved):
Thanks!

yselkowitz (changes requested):
The config needs to be split out between c9s and eln, as this change should only apply to the latter.

jkonecny12:
Is it better @yselkowitz now?

yselkowitz (approved):

PR #1027, opened 2023-11-27
Remove `numad` and `pinfo` from ELN

lzaoral:
The plan is to provide these packages in EPEL 10 instead.

\cc @lnykryn

### DIFF
--- a/configs/sst_cs_plumbers-unwanted.yaml
+++ b/configs/sst_cs_plumbers-unwanted.yaml
@@ -25,6 +25,7 @@ data:
   - libtar-devel
   - network-scripts
   - numad
+  - pinfo
   - python3-bottle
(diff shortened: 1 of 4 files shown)


configs/sst_cs_plumbers-unwanted.yaml
```diff
@@ -24,6 +24,7 @@ data:
   - libtar
   - libtar-devel
   - network-scripts
+  - numad
```
yselkowitz, line 27: Do you want to mark pinfo as explicitly unwanted here as well?

lzaoral: Addressed in 5fc82686e6b4e1d0f001a03136c80315a319badf.

PR #1029, opened 2023-11-28
s390utils is available for all arches

sharkcz:
Some of the tools from the s390utils source package are now built for all platforms, thus move the s390utils entry from the s390x section to the general one.

Related: https://issues.redhat.com/browse/RHEL-10567

### DIFF
--- a/configs/sst_arch_hw.yaml
+++ b/configs/sst_arch_hw.yaml
@@ -28,6 +28,7 @@ data:
   - hwloc
   - pigz
   - ethtool
+  - s390utils
 
   arch_packages:
     ppc64le:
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):
Yay, finally!

yselkowitz (changes requested):
This config is used for both c9s and eln, but (at least so far) the multi-arch build changes are only in ELN.  Unless that is going to be backported to c9s, then the configs will need to be split.

sharkcz:
> This config is used for both c9s and eln, but (at least so far) the multi-arch build changes are only in ELN. Unless that is going to be backported to c9s, then the configs will need to be split.

We are enabling the all-arch builds for RHEL as well, please see the linked Jira ticket.

yselkowitz (approved):

PR #1030, opened 2023-11-28
Ignore java-*-openjdk-portable in ELN

yselkowitz:
java-openjdk-portable packages are built differently internally and therefore not intended to be imported into the next RHEL version.  With Fedora's build now using pandoc, which is unwanted in RHEL due to its GHC dependencies, to build documentation, we need to ignore it entirely.

/cc @asamalik


asamalik:
Happy to try that out!

sgallagher:
Let's see how it goes!

yselkowitz:
This wasn't enough; filed #1031.

PR #1032, opened 2023-11-29
Add GNOME Clocks in RHEL 10

tpopela:
Without it it's impossible to add alternative time zones in GNOME.

See https://issues.redhat.com/browse/RHEL-4230

### DIFF
--- a/configs/sst_desktop-gnome-clocks.yaml
+++ b/configs/sst_desktop-gnome-clocks.yaml
@@ -1,0 +1,12 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: GNOME Clocks
+  description: Clocks application for Workstation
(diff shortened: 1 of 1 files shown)


yselkowitz (approved):

Conan-Kudo:
Can we please consider consolidating all these configs for the GNOME desktop? It doesn't make a lot of sense to have a yaml file per app when it's part of a logical collection.

bstinsonmhk:
We may change the organization of these in the future. For now it's up to the individual teams to choose the structure that makes the most sense for them

tpopela:
@Conan-Kudo that will be really hard, because the GNOME related components are maintained by 4 different subsystems and the ownership has to be clearly defined.

PR #1038, opened 2023-12-07
Test potential bootc changes

smooge:
- Add in initoverlayfs
- Test to see if c9s is possible or not.


cgwalters (approved):

cgwalters:
Sorry I've now reverted this in https://github.com/minimization/content-resolver-input/pull/1172

At the time, my thought was that it made sense to have "things related to boot outside systemd" in sst_bootc but this one is not in the critical path for us and given team composition and resourcing I don't want our SST to maintain it.

PR #1040, opened 2023-12-12
Update .NET workload for CentOS Stream 10

omajid:
- Add .NET 8
- Drop .NET 6, and 7

### DIFF
--- a/configs/sst_dotnet-unwanted-eln.yaml
+++ b/configs/sst_dotnet-unwanted-eln.yaml
@@ -6,7 +6,6 @@ data:
   maintainer: sst_dotnet
   labels:
   - eln
-  - c9s
 
   unwanted_packages:
   # We do not want to ship .NET Core 3.1 which will be EOL by late 2022
@@ -17,3 +16,11 @@ data:
   - dotnet-sdk-5.0
   - dotnet-runtime-5.0
   - aspnetcore-runtime-5.0
+  # We do not want to ship .NET 6.0 which will be EOL by late 2024
+  - dotnet-sdk-6.0
+  - dotnet-runtime-6.0
+  - aspnetcore-runtime-6.0
+  # We do not want to ship .NET 7.0 which will be EOL by mid 2024
+  - dotnet-sdk-7.0
+  - dotnet-runtime-7.0
+  - aspnetcore-runtime-7.0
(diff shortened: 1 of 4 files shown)


configs/sst_dotnet-dotnet.yaml
```diff
@@ -2,24 +2,14 @@ document: feedback-pipeline-workload
 version: 1
 data:
   name: .NET
-  description: .NET and .NET Core SDK, tools and runtime
+  description: .NET SDK, tools and runtime
   maintainer: sst_dotnet
 
-  packages: []
-  
-  arch_packages:
```
omajid, line 10: .NET 7 and later are available for all 4 architectures that Fedora/RHEL care about.

yselkowitz (changes requested):
These config files are currently tagged both `eln` and `c9s`.  Therefore, the configs must be split, with the removals applying only to the eln config.

omajid:
> These config files are currently tagged both eln and c9s. Therefore, the configs must be split, with the removals applying only to the eln config.

Do we care about `c9s` in the context of minimization? Should I just remove `c9s` instead?

yselkowitz:
> Do we care about `c9s` in the context of minimization? Should I just remove `c9s` instead?

We are continuing to track c9s in CR at this time.  As these removals must only apply to ELN, the configs must be split, with the removals applying only to the eln config.  Of course, if/when dotnet8.0 is added to c9s as well, then it should also be added to the c9s configs, but the previous version, having already been shipped, must not be removed.

omajid:
Thanks for clarifying that. I have split up eln and c9s configuration.

configs/sst_dotnet-unwanted-eln.yaml
```diff
@@ -0,0 +1,26 @@
+document: feedback-pipeline-unwanted
+version: 1
+data:
+  name: .NET Unwanted Packages
+  description: Packages that .NET maintainers do not wish to maintain in RHEL 9
```
yselkowitz, line 5:
```suggestion
  description: Packages that .NET maintainers do not wish to maintain in RHEL 10
```

yselkowitz (changes requested):
One small nitpick, looks good otherwise.

yselkowitz (approved):

PR #1041, opened 2023-12-13
Update fonts package list as per current lorax runtime-install.tmpl file

pnemade:
I found that sst_program-lorax-template-eln.yaml needs some update for fonts package list as per written in https://github.com/weldr/lorax/blob/master/share/templates.d/99-generic/runtime-install.tmpl file.

This PR will update the fonts package list.
@bstinsonmhk I see you are maintainer of this lorax template packages workload. 
Can you please review and merge this PR?
Thanks


pnemade:
@yselkowitz Can you please review and merge this PR?

PR #1045, opened 2024-01-03
Add Anaconda Initial Setup to ELN Extras

Conan-Kudo:
Various Enterprise Linux setups will want to be able to do this for things like spins/remixes or mass deployments.


Conan-Kudo:
FYI @tdawson @jkonecny12

PR #1047, opened 2024-01-05
ELN: enhance unwanted repository filtering

yselkowitz:
We do not need to filter exim, esmtp, etc. here because we use hints in fedora-release to choose postfix as the provider of smtpdaemon.  But we can use this filtering to block certain unwanted packages which are frequent culprits in pulling in hundreds or even a thousand-some more unwanted packages into ELN.

### DIFF
--- a/configs/repository-fedora-eln.yaml
+++ b/configs/repository-fedora-eln.yaml
@@ -11,32 +11,27 @@ data:
         baseurl: https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/compose/BaseOS/$basearch/os/
         koji_api_url: https://koji.fedoraproject.org/kojihub
         koji_files_url: https://kojipkgs.fedoraproject.org
-        exclude: ["exim", "esmtp", "opensmtpd", "esmtp-local-delivery"]
         priority: 1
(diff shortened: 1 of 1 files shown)


tdawson:
I really think you want to keep excluding exim, esmtp, etc..  Sure fedora-release might give a hint, but I'd feel better keeping them in incase dnf can't take a hint. ;)  And I don't see the harm in keeping them in.

yselkowitz:
> I really think you want to keep excluding exim, esmtp, etc.. Sure fedora-release might give a hint, but I'd feel better keeping them in incase dnf can't take a hint. ;) And I don't see the harm in keeping them in.

AFAICS it was an attempt to do what is better accomplished through hints, and since those were added to fedora-release, there have been no warnings related to these packages.  At this point they would only show up if explicitly required, and would be no different than any other unwanted package.

tdawson:
I don't have a strong opinion on exim, esmtp, etc. so if you think it's ok to take them out of excludes, I'm fine with that.
Other than that, LGTM.

yselkowitz:
Package list is sane again, so let's try to hold off for now.

PR #1048, opened 2024-01-05
remove tex-fonts-hebrew from ELN branch

ngothan:
please remove  tex-fonts-hebrew from ELN branch
Thanks!


yselkowitz (changes requested):
Please sync your local branch, there should only be one commit for this.

PR #1049, opened 2024-01-06
add bash-color-prompt

juhp:
New default package in F39


yselkowitz:
Are you planning on adding this to RHEL 9 as well, or just RHEL 10?  If the latter, this config needs to be split.

juhp:
Ah thank you - good point - just RHEL 10 indeed

PR #1053, opened 2024-01-10
Add efs-utils-selinux subpackage

major:
Once efs-utils was packaged, we realized that a SELinux policy subpackage would be required.


FKolwa (approved):
I confirm Major's observation in regard to the missing selinux package.

yselkowitz (approved):

major:
Thank you, @FKolwa and @yselkowitz! 👏

PR #1054, opened 2024-01-10
Update ELN10 NBDE related components

sergio-correia:
Replace http-parser[-devel] with llhttp[-devel]

### DIFF
--- a/configs/sst_security_special_projects-nbde.yaml
+++ b/configs/sst_security_special_projects-nbde.yaml
@@ -11,8 +11,8 @@ data:
   - clevis-luks
   - clevis-systemd
   - clevis-udisks2
-  - http-parser
-  - http-parser-devel
+  - llhttp
+  - llhttp-devel
   - jose
   - libjose
   - libjose-devel


yselkowitz:
For the record, the current consumers of http-parser in ELN have yet to be converted to llhttp:
libgit2: https://github.com/libgit2/libgit2/discussions/5828
tang: https://github.com/latchset/tang/pull/136 (from OP)

"Someone" will need to do the work in libgit2 as well in order for us to be able to drop libgit2 from ELN (and RHEL 10).  Thank you though for the advance warning on llhttp.

sgallagher:
FYI, I've got a draft started for libgit2 [here](https://github.com/libgit2/libgit2/pull/6713), but I'm not sure when I'll have the cycles to drive it to completion.

sergio-correia:
> FYI, I've got a draft started for libgit2 [here](https://github.com/libgit2/libgit2/pull/6713), but I'm not sure when I'll have the cycles to drive it to completion.

I forked your branch and did some minor fixes in this branch: https://github.com/sergio-correia/libgit2/tree/llhttp

I also changed it to try to use system libs first, initially try llhttp, then http-parser, and finally fall back to the bundled http-parser. I then added a container based on rawhide, to test llhttp. Not sure these + the other CI changes are correct, but maybe you can update your PR so we can check it out.

PR #1058, opened 2024-01-15
Add qt6-qtspeech-speechd

yselkowitz:
This is the supported backend for RHEL, but was split out of the main package to better handle other use cases.

/cc @tpopela


tpopela:
@yselkowitz LGTM!

PR #1059, opened 2024-01-15
Drop java-17-openjdk from ELN

gnu-andrew:
The OpenJDK team only plan to support java-21-openjdk in the next release.

### DIFF
--- a/configs/sst_java-java.yaml
+++ b/configs/sst_java-java.yaml
@@ -8,24 +8,24 @@ data:
   - eln
 
   packages:
-  - java-17-openjdk
-  - java-17-openjdk-demo
-  - java-17-openjdk-devel
-  - java-17-openjdk-headless
-  - java-17-openjdk-javadoc
-  - java-17-openjdk-javadoc-zip
-  - java-17-openjdk-jmods
-  - java-17-openjdk-src
-  - java-17-openjdk-static-libs
+  - java-21-openjdk
+  - java-21-openjdk-demo
+  - java-21-openjdk-devel
+  - java-21-openjdk-headless
+  - java-21-openjdk-javadoc
+  - java-21-openjdk-javadoc-zip
+  - java-21-openjdk-jmods
+  - java-21-openjdk-src
+  - java-21-openjdk-static-libs
   - copy-jdk-configs
   # Debug packages should be built but shipped in CRB
-  - java-17-openjdk-slowdebug
-  - java-17-openjdk-demo-slowdebug
-  - java-17-openjdk-devel-slowdebug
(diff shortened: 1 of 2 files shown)


yselkowitz:
Thanks for the heads-up, but this is premature.  We don't have java-21-openjdk yet at all in rawhide or ELN (java-latest doesn't count), and once we do, we'll need those packages enumerated in CR.  I'll mark this as draft in the meantime.

gnu-andrew:
RHEL 8 & 9 already have `java-21-openjdk` and RHEL 10 will be the same as those branches. We don't intend to use anything from Fedora.

yselkowitz:
We still need 21 in ELN, and everything switched to it, before we can drop 17 from ELN.  So, we're blocked by 1) needing a java-21-openjdk in rawhide, and 2) having the java-21 packages listed in Content Resolver (instead of the placeholder currently there).

gnu-andrew:
Ok, I'm just concerned that `java-17-openjdk` packages are currently being built that are not only unwanted, but also out of sync with the ones in other RHEL versions and pulling in Fedora packages that won't be available to RHEL e.g. `java-17-openjdk-portable-devel-fastdebug-1:17.0.9.0.9-4.fc38.x86_64`

yselkowitz:
java-*-openjdk-portable are explicitly blocked from ELN and from being synced to RHEL.  While that does mean that ELN java-*-openjdk is not exactly identical to RHEL's (and therefore they will need to be changed in c10s post-branching), it's the closest we can get, and do need an openjdk in ELN to build other Java packages in ELN.

gnu-andrew:
So it's expected that ELN uses Fedora packages? That surprised me.

Is there a date for c10s becoming independent of ELN? 

I saw that a `java-21-openjdk-portable` has been created in Fedora today, so hopefully `java-21-openjdk` won't be as far away as I thought and we can resolve this.

judovana:
https://fedoraproject.org/wiki/Changes/Java21#java-21-openjdk_as_the_system_JDK_in_F40 yes it will be.

yselkowitz:
There is a java-21-openjdk build in rawhide and ELN now, so this can be adjusted to drop the 21 placeholder and list 21's packages instead of 17's.

yselkowitz:
And at what point will everything be built against 21 in rawhide?

judovana:
We are working on it. eta aprox month, maybe two.

gnu-andrew:
> There is a java-21-openjdk build in rawhide and ELN now, so this can be adjusted to drop the 21 placeholder and list 21's packages instead of 17's.

Done.

PR #1062, opened 2024-01-21
Drop cyrus-sasl-ntlm from ELN

yselkowitz:
The NTLM plugin has been disabled in rawhide in preparation for it being removed upstream in a future version:

https://github.com/cyrusimap/cyrus-sasl/issues/708
https://src.fedoraproject.org/rpms/cyrus-sasl/c/8a986014fe89ef82e851d5744662e201f515e146

/cc @rcritten


rcritten:
I'm not very familiar with these files but if I'm reading it correctly it is separating c9s which will contain ntlm and eln which will not. If that's the case then LGTM.

PR #1068, opened 2024-01-29
add Secure Execution data to s390utils

sharkcz:
The new s390utils-se-data subpackage contains noarch data for Secure Execution on the s390x platform.

Related: https://issues.redhat.com/browse/RHEL-10567


sharkcz:
The subpackage needs to exist on all platforms as it complements the tools in the s390utils rpm added in PR #1029.

PR #1073, opened 2024-01-29
Add openscap-report to ELN

Mab879:
Add a new workload to add the `openscap-report` package to ELN.

### DIFF
--- a/configs/sst_security_compliance-reporting.yaml
+++ b/configs/sst_security_compliance-reporting.yaml
@@ -1,0 +1,12 @@
+document: feedback-pipeline-workload
+version: 1
+data:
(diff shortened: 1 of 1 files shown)


yselkowitz:
Thank you for the patch.  Before this package can be added to ELN, its dependencies need to be fixed to avoid adding unwanted packages to the buildroot.  These have been filed upstream as https://github.com/OpenSCAP/openscap-report/pull/225 and downstream as https://src.fedoraproject.org/rpms/openscap-report/pull-request/16.  Once fixed downstream, we should be able to proceed.

yselkowitz:
Dependencies look much better now, merging.

Issue #861, opened 2023-06-16
Update audio playback workload for ELN/c10s?


yselkowitz:
@csoriano1618 a couple questions regarding [the audio playback workload](https://github.com/minimization/content-resolver-input/blob/main/configs/sst_gpu-audio-playback.yaml):
-  can openal-soft be dropped from ELN?  qt5 is being replaced by qt6, which only uses it on the WebAssembly platform, so after https://src.fedoraproject.org/rpms/qt6-qtmultimedia/pull-request/2 nothing in ELN will depend upon it.
- Is the pulseaudio daemon still wanted in RHEL 10, or should `pulseaudio` be switched to `pipewire-pulseaudio`?

yselkowitz:
ping @csoriano1618

PR #1074, opened 2024-01-30
eln: Replace pulseaudio package with pipewire-pulse

nielsdg:
Pipewire has an equivalent implementation of PulseAudio's APIs, provided by `pipewire-pulse`. With Pipewire 1.0 being released and included, we feel confident that Pipewire now works well enough to remove the PulseAudio daemon itself.

In the very worst case, we can still add it back if needed.

### DIFF
--- a/configs/sst_gpu-audio-playback-eln.yaml
+++ b/configs/sst_gpu-audio-playback-eln.yaml
@@ -1,0 +1,15 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Audio playback
+  description: Libraries and daemons for audio playback
+  maintainer: sst_gpu
+
+  packages:
+    - pipewire-pulseaudio
+    - libcanberra
+    - openal-soft
+    - pavucontrol
+
+  labels:
+    - eln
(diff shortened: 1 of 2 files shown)


nielsdg:
Related: https://github.com/minimization/content-resolver-input/issues/861

configs/sst_gpu-audio-playback-eln.yaml
```diff
@@ -0,0 +1,15 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Audio playback
+  description: Libraries and daemons for audio playback
+  maintainer: sst_gpu
+
+  packages:
+    - pipewire-pulse
```
yselkowitz, line 9:
```suggestion
    - pipewire-pulseaudio
```

yselkowitz (changes requested):

configs/sst_gpu-audio-playback-eln.yaml
nielsdg: eek, you're right. Fixed now!

yselkowitz (approved):

PR #1078, opened 2024-02-01
[sst_kernel_rts] Remove eln from labels in sysfsutils

wchwhite:
Removing eln from the labels, but leaving cs9. Sysfsutils is being removed from RHEL10.


yselkowitz:
@wchwhite please note that sblim-gather still has a dependency on libsysfs, so sysfsutils will not be dropped entirely until that changes.

PR #1079, opened 2024-02-02
eln: Remove openal-soft from audio workload

nielsdg:
It seems it was added as a dependency of SDL_sound and Qt5, so we should be good to remove this. If need be, we can always put it back.

Fixes: https://github.com/minimization/content-resolver-input/issues/861


yselkowitz:
Thanks, nothing in ELN currently requires openal-soft, so this should be a simple drop.

PR #1084, opened 2024-02-06
Remove `gpm` and its subpackages from ELN

pcahyna:
gpm provides support for using mouse on Linux text console, which is IMO a rarely used functionality and not worth maintaining.

Also split cs_system_management ELN and C9S workloads. sst_cs_system_management-sys-management.yaml had both `c9s` and `eln` labels and sst_cs_system_management-sys-management-c9s.yaml contained only the packages that are only in C9S. Make the workloads independent by copying everything from sst_cs_system_management-sys-management.yaml to sst_cs_system_management-sys-management-c9s.yaml and removing the `c9s` tag from the former. Since C9S has already branched from ELN and is independent, having separate workloads will be easier to manage than sharing the common content.

### DIFF
--- a/configs/sst_cs_system_management-sys-management.yaml
+++ b/configs/sst_cs_system_management-sys-management.yaml
@@ -123,4 +123,3 @@ data:
 
   labels:
   - eln
-  - c9s
--- a/configs/sst_cs_system_management-sys-management.yaml
+++ b/configs/sst_cs_system_management-sys-management.yaml
@@ -17,9 +17,6 @@ data:
   - eigen3-devel
   - freeipmi
   - freeipmi-devel
-  - gpm
-  - gpm-devel
-  - gpm-libs
   - ipmitool
   - iprutils
   - jasper-devel
--- a/configs/sst_cs_system_management-sys-management-c9s.yaml
+++ b/configs/sst_cs_system_management-sys-management-c9s.yaml
@@ -6,12 +6,123 @@ data:
   maintainer: sst_cs_system_management
 
   packages:
-  - modulemd-tools
+  - babl-devel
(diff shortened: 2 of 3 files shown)


yselkowitz:
emacs and mc currently depend on gpm, is the intention that they should no longer do so in ELN or RHEL 10?

pcahyna:
Hello @yselkowitz the intent is to drop the dependency, it is not necessary for the packages' functionality. https://issues.redhat.com/browse/RHEL-23695 https://issues.redhat.com/browse/RHEL-23700

configs/sst_cs_system_management-sys-management.yaml
```diff
@@ -123,4 +123,3 @@ data:
 
   labels:
   - eln
-  - c9s
```
yselkowitz, line 126: The gpm packages should be removed from this file.

yselkowitz (changes requested):
One change is needed.  In the meantime, I have filed https://src.fedoraproject.org/rpms/emacs/pull-request/38 and https://src.fedoraproject.org/rpms/mc/pull-request/3 to drop the current dependencies on gpm.

configs/sst_cs_system_management-sys-management.yaml
pcahyna: thanks for noticing it, I forgot to commit my changes - corrected in the latest commit.

lzaoral:
@yselkowitz Thank you for filling out those PRs!  However, I would have appreciated to also assign/mark that progress in the linked Jira tickets so that the assigned people would not start doing something and then they would realise that it was already done.

yselkowitz (approved):

PR #1085, opened 2024-02-06
Infra Services c10s workload updates:

notroj:
- restore varnish
- drop spamassassin, sendmail
- explicitly mark unwanted removed packages

### DIFF
--- a/configs/sst_cs_infra_services-mta-c10s.yaml
+++ b/configs/sst_cs_infra_services-mta-c10s.yaml
@@ -14,10 +14,7 @@ data:
   - postfix-perl-scripts
   - postfix-pgsql
   - postfix-sqlite
-  - sendmail
-  - sendmail-cf
-  - sendmail-doc
-  - sendmail-milter
 
   labels:
-  - c9s
+  - eln
+  - c10s
--- a/configs/sst_cs_infra_services-spam-filtering.yaml
+++ b/configs/sst_cs_infra_services-spam-filtering.yaml
@@ -9,5 +9,4 @@ data:
   - spamassassin
 
   labels:
-  - eln
   - c9s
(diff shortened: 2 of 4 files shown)


yselkowitz:
@notroj just noticed now that sudo has a hard `BuildRequires: sendmail`.  Is there any reason why it could not `BuildRequires: server(smtp)` or `/usr/sbin/sendmail` instead, and use whichever provider the user chooses (postfix for RHEL 10, but Fedora has lots of choices) at runtime, or is sudo really only compatible with sendmail?

yselkowitz:
Actually, looking closer at sudo's sources, it looks like we could just configure --with-sendmail=/usr/sbin/sendmail and not BuildRequires it at all?

yselkowitz:
> Actually, looking closer at sudo's sources, it looks like we could just configure --with-sendmail=/usr/sbin/sendmail and not BuildRequires it at all?

https://src.fedoraproject.org/rpms/sudo/pull-request/27

notroj:
A big thank you @yselkowitz 

```
+ dnf repoquery -q --whatdepends /usr/sbin/sendmail
redhat-lsb-core-0:4.1-56.el9.i686
redhat-lsb-core-0:4.1-56.el9.x86_64
```
```
+ dnf repoquery -q --whatdepends sendmail
e2scrub-0:1.46.5-5.el9.x86_64
...
```

I guess the first one is OK, but not the second one. I see there is already a PR for this too

https://src.fedoraproject.org/rpms/e2fsprogs/pull-request/6

yselkowitz:
Actually, neither show up as affecting ELN: https://tiny.distro.builders/view-rpm--view-eln--sendmail.html (which will go away once the sudo PR is merged and goes through the pipeline).  redhat-lsb-core has been removed from RHEL, and e2scrub (subpackage of e2fsprogs) is built but not shipped in ELN or RHEL either.  It still wouldn't hurt to change the hardcoded `sendmail` dependency to something more generic though, even if "just" for Fedora's benefit.

PR #1086, opened 2024-02-07
Drop sendmail-milter-devel from ELN

yselkowitz:
This is a follow-up to #1085 which dropped the rest of sendmail.

/cc @notroj


notroj:
LGTM, thanks @yselkowitz

PR #1087, opened 2024-02-08
Add perl-XML-Simple to ELN

jplesnik:
It was required in JIRA RHEL-24166


jplesnik:
The package was already in RHEL 9 ([RHELPLAN-92234](https://issues.redhat.com/browse/RHELPLAN-92234))

PR #1092, opened 2024-02-12
Add intel-lpmd package

Jiri-Koten:
Add intel-lpmd package to enable Low Power Mode on Intel Meteor Lake platforms.
See https://issues.redhat.com/browse/RHEL-20167.


Jiri-Koten:
There is a plan to  migrate from dbus-glib to GDBus and we are targeting RHEL 10 GA.

PR #1093, opened 2024-02-12
[sst_cloud_experience] Add opentelemetry-collector

major:
Adding a package for the opentelemetry collector as well as AWS Cloudwatch configuration files.

### DIFF
--- a/configs/sst_cloud_experience.yaml
+++ b/configs/sst_cloud_experience.yaml
@@ -25,6 +25,17 @@ data:
       - awscli2
       - python3-awscrt
 
+  package_placeholders:
+    - srpm_name: opentelemetry-collector
+      rpms:
+        - rpm_name: opentelemetry-collector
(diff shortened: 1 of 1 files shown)


yselkowitz:
@major I can't find such a package in Fedora, nor any review requests for it to be added.  Is this going to be a RHEL-specific package?  If so, it needs to be handled as a placeholder.

major:
Thanks for catching that, @yselkowitz. I've submitted a new patch.

yselkowitz:
Thanks for clarifying.  That placeholder syntax is deprecated, but unfortunately still commonly used as we haven't updated all uses of it in CR.  (Another thing we need to look at after RHEL 10 branching.)  Please use this as an example of the new syntax: https://github.com/minimization/content-resolver-input/blob/main/configs/sst_insights.yaml#L12

major:
Thank you! I've just updated the patch to follow the new syntax.

PR #1096, opened 2024-02-13
For CS10 add tuned-ppd, replacing power-profiles-daemon

notroj:
CC @yarda @Jiri-Koten


yselkowitz:
ELN is currently in a freeze for the F40 and c10s branching efforts.  We'll be able to take a look at this after the freeze lifts.

PR #1098, opened 2024-02-14
Add c10s to Content Resolver

yselkowitz:
Please check this carefully, let me know if anything is missing, and we need to decide when to enable this.

The first commit only affects whitespace, to make the following commit easier to review.


sgallagher (approved):
It looks reasonable to me. I don't see a particular reason to delay merging it, other than that it will double the time between runs of Content Resolver.

yselkowitz:
c10s has branched, so tracking that is a higher priority than ELN frequency, which is much less urgent now.  I wouldn't mind @asamalik looking this over first though.

asamalik:
I haven't reviewed the _Reformat with yq_ commit, I assume it's just (well needed) a formatting change.

Otherwise looks good to me! 
- You've added the repo, the environment, and the view, all correctly linked together and using the `c10s` tag
- You've added the new `c10s` tag to everything having the `eln` tag
- You haven't added the new `c10s` tag to the `c9s`-only workloads, which is correct, because we're branching from ELN

Can't think of anything else. Now let's see if the builder has enough capacity to process all this.

asamalik (approved):

PR #1099, opened 2024-02-16
Add pcre2-tools as for the customer RFE

ljavorsk:
Related: RHEL-23512


yselkowitz (approved):
Straightforward addition of a subpackage of an existing package.  Confirmed addition to both 9 and 10.

PR #1100, opened 2024-02-21
Add libzip-tools as for the customer RFE

ljavorsk:
Related: RHEL-17567


yselkowitz (approved):
This is just an additional subpackage from an existing package, so no further steps required for c10s inclusion.

PR #1101, opened 2024-02-21
Make scap-security-guide a placeholder

yselkowitz:
scap-security-guide in Fedora includes all OS data, but in RHEL it only includes data for that RHEL version.  Since that RHEL version specific data needs to be created as part of development, it cannot be so built in ELN or imported when RHEL branches.  Also, there is a RHEL-specific rule-playbooks subpackage.  As such, scap-security-guide fits the criteria of a placeholder, and handling it as such avoids issues.


yselkowitz:
ping @bstinsonmhk @sgallagher

sgallagher:
This is necessary but not sufficient. There needs to be a corresponding MR for the [`distrobaker.yaml`](https://gitlab.com/redhat/centos-stream/ci-cd/distrosync/distrobuildsync-config/-/blob/main/distrobaker.yaml?ref_type=heads) to exclude it from ELN rebuilds since placeholders still appear on the list downloads. See for example `fence-agents` in https://tiny.distro.builders/view-all-source-package-name-list--view-eln.txt

yselkowitz:
> This is necessary but not sufficient. There needs to be a corresponding MR for the [`distrobaker.yaml`](https://gitlab.com/redhat/centos-stream/ci-cd/distrosync/distrobuildsync-config/-/blob/main/distrobaker.yaml?ref_type=heads) to exclude it from ELN rebuilds since placeholders still appear on the list downloads. See for example `fence-agents` in https://tiny.distro.builders/view-all-source-package-name-list--view-eln.txt

All placeholders appear on that list, yet we haven't excluded any of them (or more specifically, any of those that actually exist in Fedora) from distrobuildsync?

PR #1102, opened 2024-02-28
Drop NetworkManager-team subpackage in ELN and c10s

smwfaye:
Drop NetworkManager-team subpackage in ELN and c10s.

### DIFF
--- a/configs/sst_network_management-unwanted-c10s.yaml
+++ b/configs/sst_network_management-unwanted-c10s.yaml
@@ -1,0 +1,13 @@
+document: feedback-pipeline-unwanted
+version: 1
+data:
+  name: Network Management unwanted
+  description: >
+    Unwanted packages from Network Management
+
+  maintainer: sst_network_management
+  unwanted_packages:
+    - NetworkManager-team
+  labels:
+    - eln
+    - c10s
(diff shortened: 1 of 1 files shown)


yselkowitz:
OK, but this won't actually drop it until all dependencies thereon are removed.  Namely, from anaconda: https://github.com/rhinstaller/anaconda/blob/master/anaconda.spec.in#L121 so please coordinate that with the installer team.

Furthermore, does that means that libteam/teamd should also be dropped from ELN and c10s?  The only dependents are anaconda and NM-team.  If so, then the `team` bcond in NetworkManager.spec should be disabled for RHEL 10+ once the anaconda dependencies are dropped.

yselkowitz:
ping @smwfaye ?

configs/sst_network_management-unwanted-c10s.yaml
```diff
@@ -0,0 +1,13 @@
+document: feedback-pipeline-unwanted
+version: 1
+data:
+  name: Network Management unwanted
+  description: >
+    Unwanted packages from Network Management
+
+  maintainer: sst_network_management
+  unwanted_packages:
+    - NetworkManager-team
```
yselkowitz, line 10:
```suggestion
    - NetworkManager-team
    - libteam
    - teamd
    - teamd-devel
```

yselkowitz (changes requested):
In that case, let's go ahead and mark teamd packages as unwanted now.

yselkowitz (approved):

PR #1105, opened 2024-03-04
Add DNF 5 to RHEL 11

sgallagher:
Part of https://github.com/fedora-eln/eln/issues/175


sgallagher:
Yeah, right now we still expect to use DNF 4 at runtime, but DNF 5 at build-time.

yselkowitz (approved):

PR #1108, opened 2024-03-07
Fixup the Python devel (CRB) workload for c10s

hroncok:
sst_cs_apps-python-crb-c10s.yaml is a copy of sst_cs_apps-python-crb-eln.yaml.

 - labels were adjusted (split)
 - python-wheel-wheel is called python3-wheel-wheel in c10s

### DIFF
--- a/configs/sst_cs_apps-python-crb-c10s.yaml
+++ b/configs/sst_cs_apps-python-crb-c10s.yaml
@@ -18,8 +18,8 @@ data:
     - python3-cython
     - python3-sphinx
     - python3-wheel
-    # note: the following package is called python3-wheel-wheel in RHEL proper:
-    - python-wheel-wheel
+    # note: the following package is called python-wheel-wheel in ELN:
+    - python3-wheel-wheel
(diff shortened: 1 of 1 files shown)


yselkowitz:
Could we fix this naming discrepancy in ELN instead?

hroncok:
Maybe. But it's complicated. The package is named after the path and the path is different based on a macro value that is different in ELN and RHEL. The difference was approved in https://github.com/fedora-eln/eln/issues/73

If we do change things, it won't be today. Please don't block merging this on that.

hroncok:
If you don't like splitting the files, we can remove the line instead, as the package is now required by python3-test anyway. But it was included in c9s CRB explicitly before that dependency existed.

yselkowitz:
It's not about splitting the files as it is with the issues this discrepancy causes when bootstrapping a new version of CS and RHEL.  Let's follow up in https://github.com/fedora-eln/eln/issues/181 (or in a new issue spun out from that).

PR #1110, opened 2024-03-13
Revert disabling numad

psimovec:


yselkowitz (approved):
numad is a dependency of libvirt-daemon-driver-qemu, so this reflects reality.

PR #1113, opened 2024-03-19
Drop sushi from c10s and eln

yselkowitz:
Resolves: https://issues.redhat.com/browse/CS-2020

/cc @tpopela


tpopela:
Thank you @yselkowitz !

PR #1119, opened 2024-03-26
Update ELN Extras workload for Meta

davide125:
Add a couple more packages we need.

### DIFF
--- a/configs/eln_extras_meta.yaml
+++ b/configs/eln_extras_meta.yaml
@@ -46,6 +46,7 @@ data:
     - libkdumpfile-devel
     - libidn
     - libidn-devel
+    - linux-sysinfo-snapshot
     - mlnx-tools
     - mkosi
     - mosh


configs/eln_extras_meta.yaml
```diff
@@ -27,6 +27,7 @@ data:
     - duo_unix
     - duo_unix-selinux
     - ebranch
+    - emacs-nw
```
yselkowitz, line 30: emacs-nw is in ELN.

davide125: Is it actually shipped though? `mock -r fedora-eln-x86_64 --install emacs-nw` fails for me here.

davide125: I see it listed in https://github.com/minimization/content-resolver-input/blob/main/configs/sst_cs_plumbers-editors.yaml so in theory it should be (and https://tiny.distro.builders/workload--sst_cs_plumbers-editors--fedora-empty-base--repository-fedora-eln--x86_64.html confirms it), but I don't see in https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/compose/AppStream/x86_64/os/Packages/ .

yselkowitz: That would need to be fixed in https://pagure.io/pungi-fedora/tree/eln , not in content-resolver.

davide125: Pungi is just installing the emacs group, so I think the actual fix for this needs to be in comps: https://pagure.io/fedora-comps/pull-request/950

yselkowitz (approved):

PR #1122, opened 2024-04-03
Add FreeRADIUS subpackages

antoniotorresm:
These subpackages should be checked as well when running content resolver.


yselkowitz (approved):
Confirmed that these are already in RHEL9.

antoniotorresm:
@yselkowitz we are in the process of removing this dependency, so no issues. Thanks!

PR #1123, opened 2024-04-04
Drop awscli from ELN Extras for Meta

yselkowitz:
The awscli2 package has replaced and provides awscli, and is in ELN. Furthermore, while noarch, it cannot be installed on s390x due to arch limiations in python-awscrt (rhbz#2180988).  Therefore, if relisted in the future, it must be limited to installable architectures in the workload.

/cc @davide125

### DIFF
--- a/configs/eln_extras_meta.yaml
+++ b/configs/eln_extras_meta.yaml
@@ -6,7 +6,6 @@ data:
   maintainer: meta-sig
   packages:
     - atop
-    - awscli
     - azure-cli
     - b4
     - below


michel-slm:
This is likely fine, but we need to follow up with our Cloud team to be sure. We did provide an awscli2 build in Hyperscale (since awscli2 is only in c10s and not in c9s) so hopefully we can confirm soon that we no longer need awscli 1.x

yselkowitz:
awscli 1.x was retired from Fedora, so there is no way for it to be in ELN Extras.

davide125:
Oh I'd missed this was retired. Yeah let's kill it then and we'll deal with the fallout on our end. Thanks!

davide125 (approved):

PR #1124, opened 2024-04-04
eln_extras_meta: add libxcrypt-compat

michel-slm:
This is no longer built as part of libxcrypt in RHEL >= 10 but is still needed by Chef, so it is now built from a separate libxcrypt-epel package:

https://koji.fedoraproject.org/koji/packageinfo?packageID=39931

### DIFF
--- a/configs/eln_extras_meta.yaml
+++ b/configs/eln_extras_meta.yaml
@@ -45,6 +45,7 @@ data:
     - libkdumpfile-devel
     - libidn
     - libidn-devel
+    - libxcrypt-compat
     - linux-sysinfo-snapshot
     - mlnx-tools
     - mkosi


yselkowitz:
We're going to have to adjust the automation first to not try to build this, since it's an EPEL-only package (similar to an ELN branched package).

michel-slm:
@yselkowitz there is no rawhide branch for this package - and the build is made from the ELN branch right now (we might need this for epel10 as well later, but epel10 does not exist yet). What sort of automation adjustment is needed, can I help put up a PR?

sgallagher:
I missed this until it was linked in the ELN meeting today. I agree with @michel-slm: we can merge this without waiting for the matching EBS config change, because it won't be building for the `f41` tag.

That said, it would be great if @michel-slm  could also send a merge request for https://gitlab.com/redhat/centos-stream/ci-cd/distrosync/distrobuildsync-config/-/blob/main/distrobaker.yaml

sgallagher:
I went ahead and added libxcrypt-epel to the exclusion list for EBS: https://gitlab.com/redhat/centos-stream/ci-cd/distrosync/distrobuildsync-config/-/commit/f50f8d99596cee37fa2b15cd3c62f1c8e86e72fa

PR #1127, opened 2024-04-09
Adding openscap-report to RHEL8

dahaic:
Having `openscap-report` in RHEL8 will help us consolidate experience across supported version.

Related: RHEL-32208


yselkowitz:
c8s is no longer tracked in Content Resolver.

PR #1128, opened 2024-04-09
Add zenity to CR

tpopela:
Previously it was pulled by Gedit and Mutter, but we replaced Gedit with GNOME Text Editor and the dependency was dropped from Mutter, so fix this for ELN/RHEL 11 and slowly start the PRP process for RHEL 10..


yselkowitz (approved):
Looks like gedit got its revenge for being replaced.  :-)  Let's get that fixed.

PR #1132, opened 2024-04-11
This is part of Qt 6 now

tpopela:
Part of Qt 6.7

See internal to RH https://issues.redhat.com/browse/DESKTOP-882


yselkowitz:
I would feel better waiting on this until at least c10s has 6.7 with the backport.

yselkowitz (approved):
Looks like this is in c10s now; qadwaitadecorations may now be removed.

PR #1138, opened 2024-04-23
python-pam: add as requested by Anaconda team

ikerexxe:


configs/sst_idm_sssd-python-pam.yaml
```diff
@@ -0,0 +1,11 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: python-pam
+  description: Python bindings for libpam.
+  maintainer: sst_idm_sssd
+  packages:
+    - python-pam
```
yselkowitz, line 8: This needs to be the name of the binary RPM, not the SRPM, iow:
```suggestion
    - python3-pam
```

yselkowitz (changes requested):

configs/sst_idm_sssd-python-pam.yaml
ikerexxe: Updated

yselkowitz (approved):

PR #1142, opened 2024-04-26
Move scipy ownership from cs_apps to cs_plumbers

psimovec:


configs/sst_cs_plumbers-math-c9s.yaml
```diff
@@ -0,0 +1,17 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Math
+  description: Various math related libraries
```
yselkowitz, line 5: These need to be changed so as not to conflict with your existing Math workload.  Perhaps this should be renamed to ATLAS (both here and -atlas in the file name).

yselkowitz (changes requested):

lzaoral:
@psimovec `scipy` id now owned by plumbers in all releases.  Therefore, it should be added to the `c9s` yaml as well.

psimovec:
Thanks for the comments, I have simplified the changes

PR #1143, opened 2024-04-29
sst_network_drivers: remove fabtests from Appstream

michich:
JIRA: https://issues.redhat.com/browse/RHEL-12293

It pulls in python3-pytest, which we don't want to maintain.


michich:
Oh, wait a minute, I think I need to separate the rules for c9s and c10s.

michich:
I filed a new MR: [1144](https://github.com/minimization/content-resolver-input/pull/1144)

PR #1146, opened 2024-04-30
Unwanted: efs-utils + efs-utils-selinux

major:
We cannot package efs-utils and efs-utils-selinux any longer due to upstream changes.


major:
First time doing unwanted packages, so I'm not sure if I've done this correctly.

yselkowitz (changes requested):
The existing efs-utils config(s) need to be removed instead.

major:
Thanks, @yselkowitz. I must have been following some old docs.

yselkowitz (approved):

PR #1148, opened 2024-05-01
Remove mdadm from sst_arch_hw_eln.yaml

dledford:
The mdadm package is required for the storage workload (which is where it should be).  It does not need to be required by arch_hw which no longer has any control over the md raid subsystem that mdadm is used to manage.

### DIFF
--- a/configs/sst_arch_hw.yaml
+++ b/configs/sst_arch_hw.yaml
@@ -5,7 +5,6 @@ data:
   description: Packages for hardware enablement
   maintainer: sst_arch_hw
   packages:
-    - mdadm
     - nvme-cli
     - acpica-tools
     - bluez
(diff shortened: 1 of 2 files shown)


yselkowitz:
Are you still maintaining it for RHEL 9 though?  If not, https://github.com/minimization/content-resolver-input/blob/main/configs/sst_arch_hw.yaml#L8 should also be removed.

yselkowitz:
mdadm is already covered by SST Logical Storage for all of c9s/c10s/eln.  If they are the sole maintainers across the board, then let's make that reflected here while we're at it.

dledford:
Done (against my wishes).

yselkowitz (approved):

PR #1149, opened 2024-05-01
Update arch_hw config with two patches to remove unnecessary includes

dledford:
This includes two patches (one already covered under another pull request) to remove unneeded items from the arch_hw package definitions.  In both cases, the packages are also included via other mechanisms, so no changes to the actual RHEL10 package set will happen, this will just clean up who/why the packages are being included.


dledford:
Superseded by pull request #1152

PR #1152, opened 2024-05-09
Remove nvme-cli from sst_arch_hw_eln.yaml

dledford:
This package is not maintained by arch_hw.  It is required by anaconda-install-img-deps as shown by https://tiny.distro.builders/view-rpm--view-eln--nvme-cli.html and so will still be included in the install image.  If it is wanted post-install, then likely storage_ssg should pull it in as a requirement.

### DIFF
--- a/configs/sst_arch_hw_eln.yaml
+++ b/configs/sst_arch_hw_eln.yaml
@@ -5,7 +5,6 @@ data:
   description: Packages for hardware enablement
   maintainer: sst_arch_hw
   packages:
-    - nvme-cli
     - acpica-tools
     - bluez
     - alsa-firmware


yselkowitz (changes requested):
Unfortunately there is no dependency on nvme-cli in c10s, so this would cause it to fall of CR entirely, and (mistakenly) become a candidate for removal.  Let's get a new owner for this and move it to another workload instead.

While we're at it, perhaps we can get clarification on the inclusion of both nvme-cli and nvmetcli?

dledford:
> Unfortunately there is no dependency on nvme-cli in c10s, so this would cause it to fall of CR entirely, and (mistakenly) become a candidate for removal. Let's get a new owner for this and move it to another workload instead.

It's a hard runtime dependency of anaconda-install-image-deps, doesn't that keep it from falling off entirely?

> While we're at it, perhaps we can get clarification on the inclusion of both nvme-cli and nvmetcli?

I have pinged the manager of the NVME work being done about this and asked his input.  It may be that we move it to storage workload, or it may be that the nvmetcli is a replacement.  We'll wait and see.

yselkowitz:
> It's a hard runtime dependency of anaconda-install-image-deps, doesn't that keep it from falling off entirely?

Not in c10s: https://tiny.distro.builders/view-rpm--view-c10s--nvme-cli.html

tbzatek:
`nvme-cli` needs to be added to `sst_storage_io-packages.yaml` to reflect real ownership. Same should be done for `libnvme` to confirm explicit ownership.

maurizio-lombardi:
> `nvme-cli` needs to be added to `sst_storage_io-packages.yaml` to reflect real ownership. Same should be done for `libnvme` to confirm explicit ownership.

I opened a MR for that: https://github.com/minimization/content-resolver-input/pull/1159

maurizio-lombardi:
> While we're at it, perhaps we can get clarification on the inclusion of both nvme-cli and nvmetcli?

They are two completely different packages. There is no relationship between them.

yselkowitz (approved):
#1159 will keep this in the package set.

PR #1155, opened 2024-05-13
Fix Keylime workload packages

ansasaki:
Remove all dependencies that are not unique to Keylime

### DIFF
--- a/configs/sst_security_special_projects-keylime.yaml
+++ b/configs/sst_security_special_projects-keylime.yaml
@@ -8,23 +8,10 @@ data:
   packages:
     - keylime
     - keylime-agent-rust
-    - libarchive
-    - openssl
     - python3-alembic
-    - python3-cryptography
-    - python3-gpg
-    - python3-jinja2
-    - python3-lark-parser
-    - python3-packaging
-    - python3-pyasn1
-    - python3-pyasn1-modules
-    - python3-pyyaml
-    - python3-requests
+    - python3-lark
     - python3-sqlalchemy
     - python3-tornado
-    - python3-typing-extensions
-    - tpm2-tools
-    - tpm2-tss
   labels:
     - eln
     - c9s


sergio-correia (approved):

yselkowitz:
For those dependencies which are maintained by other teams (e.g. openssl) I can understand this, but for those dependencies that are unique to keylime and therefore are (presumably) maintained by the same team, I don't see how this is an improvement.

ansasaki:
> For those dependencies which are maintained by other teams (e.g. openssl) I can understand this, but for those dependencies that are unique to keylime and therefore are (presumably) maintained by the same team, I don't see how this is an improvement.

The improvement I see is a clear message on which packages are essential to us.

If in future some dependencies are dropped or replaced upstream, we can easily follow without the risk of ending up having to maintain a package that we do not use. The fact that there are packages that are brought as dependency by Keylime, and that are only used by Keylime, is circumstantial, we do not care deeply about them. Of course we will keep maintaining such packages, but it is better to clarify which are the packages that really matter to us.

yselkowitz:
@bstinsonmhk I have mixed feelings about this; thoughts?

yselkowitz:
ping @bstinsonmhk

yselkowitz:
Dependencies that are unique to keylime and IMO should remain listed for proper accounting:
```
    - python3-alembic
    - python3-lark
    - python3-sqlalchemy
    - python3-tornado
```

configs/sst_security_special_projects-keylime.yaml
```diff
@@ -8,23 +8,6 @@ data:
   packages:
     - keylime
     - keylime-agent-rust
```
yselkowitz, line 10: @bstinsonmhk has agreed that the dependencies which are unique to keylime, and therefore by definition are being maintained by the same SST, should continue to be tracked here.
```suggestion
    - keylime-agent-rust
    - python3-alembic
    - python3-lark
    - python3-sqlalchemy
    - python3-tornado
```

yselkowitz (changes requested):

configs/sst_security_special_projects-keylime.yaml
ansasaki: OK, done

yselkowitz (approved):

PR #1156, opened 2024-05-13
Mark NetworkManager-disptacher-routing-rules as unwanted for c10s

smwfaye:
Mark NetworkManager-disptacher-routing-rules as unwanted for c10s and eln as ifcfg-rh is getting removed in rhel-10+.


yselkowitz (changes requested):
This package needs to be removed instead from sst_network_management-main.yaml, which will require splitting out the c9s config (where it will remain).  There is no point in (also) listing it as unwanted since it is not a dependency that would be pulled in.

smwfaye:
Thanks, I just updated with the suggested changes.

yselkowitz (changes requested):
You still need to remove this package from sst_network_management-main.yaml.

yselkowitz (approved):

PR #1157, opened 2024-05-13
Mark openslp-server as unwanted

vcrhonek:
OpenSLP upstream is not active and the package often suffers from security issues. Some of them are even not fixable due to protocol design flaws. However, client and devel part are still needed.


yselkowitz (approved):
Nothing currently pulls this in.

PR #1162, opened 2024-05-15
eln: Mark pulseaudio daemon package as unwanted

nielsdg:
The daemon is removed in favor of Pipewire (which contains an implementation still of the PulseAudio API)


yselkowitz (approved):
OK, not that it has shown up recently, since https://src.fedoraproject.org/rpms/fedora-release/blob/rawhide/f/fedora-release.spec#_458 is in place

PR #1163, opened 2024-05-15
Update gnome-shell-extensions

yselkowitz:
Related: RHEL-36286
Related: RHEL-36288

/cc @tpopela


tpopela:
Thank you @yselkowitz !

PR #1164, opened 2024-05-16
Revert "sst_network_drivers: remove fabtests from Appstream"

yselkowitz:
fabtests needs to be moved to CRB, not removed entirely.

This reverts commit e3e56f354f687e7bb52c18119b36be0bd264aa06 from #1144 .

/cc @michich

### DIFF
--- a/configs/sst_network_drivers-appstream.yaml
+++ b/configs/sst_network_drivers-appstream.yaml
@@ -6,6 +6,7 @@ data:
   maintainer: sst_network_drivers
   packages:
     - libfabric-devel
+    - fabtests
     - openmpi
     - openmpi-devel
     - openmpi-java


hroncok:
The name of this workload is "RDMA Stack Appstream Packages". I don't think fabtests should be there.

michich:
Yeah, that will be a bit confusing, but I'm OK with it temporarily. I can rearrange the sst_network_drivers yaml files later once I have a better understanding of the Content Resolver and Pungi configs.

hroncok:
> Yeah, that will be a bit confusing, but I'm OK with it temporarily. I can rearrange the sst_network_drivers yaml files later once I have a better understanding of the Content Resolver and Pungi configs.

Considering it has been a month, I opened https://github.com/minimization/content-resolver-input/issues/1191

PR #1166, opened 2024-05-16
Consolidate to the new sst_csi_client_tools

ptoscano:
Adapt for the newly introduced sst_csi_client_tools, which is a rename of the former sst_subscription_manager:
- consolidate & rename the existing two files for sst_subscription_manager into one
- move packages maintained by sst_csi_client_tools from sst_insights to sst_csi_client_tools

Note: sst_insights is thus left empty, since it still exists and it's supposed to have packages. Adding more packages is out of the scope of these changes, and will need to be done by a different team/people.


configs/sst_insights.yaml
```diff
@@ -4,42 +4,6 @@ data:
   name: insights
   description: Insights
   maintainer: sst_insights
-  packages:
```
yselkowitz, line 7: This needs to be: `-  packages: []`

yselkowitz (changes requested):
One change needed for sst_insights.

configs/sst_insights.yaml
ptoscano: oh ok, thanks for the notice, fixing

... maybe the `test_config_files.py` ought to check for this? ;-)

yselkowitz (approved):

PR #1172, opened 2024-05-23
sst_bootc: Drop initoverlayfs

cgwalters:
At the current time the development of this is focused on Auto primarily as far as I know and the composition of the current bootc SST does not maintain it.


smooge:
I am good with this merge.

yselkowitz (approved):

PR #1189, opened 2024-06-16
Pull hatchling into CRB

hroncok:
cc @frenzymadness @torsava


frenzymadness (approved):
Make sense. Thank you!

yselkowitz (approved):

PR #1190, opened 2024-06-17
Add more Python packages to c10s CRB workflow

hroncok:
Those are intentionally not listed in the ELN workload,
as we do not want to pull them into ELN,
we just acknowledge them in CRB as they are already ours.

cc @frenzymadness @torsava

### DIFF
--- a/configs/sst_cs_apps-python-crb-c10s.yaml
+++ b/configs/sst_cs_apps-python-crb-c10s.yaml
@@ -31,5 +31,11 @@ data:
     # rather than going trough a tad tedious process to add them after GA
     - python3-installer
     - python3-pyproject-hooks
+    # extras from packages in other Python workflows
+    - micropipenv+toml
+    - python3-jinja2+i18n
+    # packages we need as build dependencies only and are willingly maintaining them in CRB
+    - python3-testpath
+    - python-testpath-doc
   labels:
     - c10s


frenzymadness:
The only package I'm not sure about is `python-testpath-doc`. Do we need it in CRB? Is the reason that the documentation should be available as well as `python-testpath` in CRB? If so, it makes sense and you can consider this approved.

hroncok:
We don't need it in CRB. Technically, we don't need any of those packages in CRB. But since we built it and it has no extra runtime deps, I see no reason to hide it from the world.

frenzymadness (approved):

yselkowitz:
So you intend to ship these in RHEL 10 but not RHEL 11?

hroncok:
We intent to ship them in RHEL 10 but we don't want to be assigned owners of the components in RHEL 11 unless we are for other reasons. Considering content resolver is used to blame ownership, I'd rather commit this again after 3 years.

hroncok:
I assume there is no way to express "put those in CRB if we maintain them but do not make us as the implicit owners"

yselkowitz:
Can't say for sure, but I don't think you would necessarily be assigned ownership of packages which are extras from packages in other workloads.  As for testpath, given the dependents, wouldn't you be the maintainers anyway?

hroncok:
Yes, if the dependency stays. But consider our package drops the dependency next year. That way, we will not want this in ELN at all.

yselkowitz:
Teams are expected to review their workloads prior to branching, so it's not as if there won't be a chance to drop these in that case prior to c11s.

hroncok:
Ack. I would still prefer to merge this. Note that the packgaes are on their way to c10s CRB already, so this merely makes the data here reflect the reality.

PR #1192, opened 2024-06-18
Drop i686 and armv7hl

yselkowitz:
C10S and ELN are dropping all i686 builds, and even in C9S i686 was built only for multilibs and not separately composed.  32-bit ARM was never a supported platform in RHEL or CentOS Stream.

/cc @tdawson @sgallagher


tdawson:
LGTM
Thank you very much got this.

PR #1193, opened 2024-06-19
Drop marshalparser from the Python CRB ELN workload

hroncok:
It is no longer necessary in RHEL 11+ due to
https://fedoraproject.org/wiki/Changes/ReproduciblePackageBuilds

See also https://pagure.io/packaging-committee/pull-request/1371

cc @frenzymadness @torsava @keszybz

### DIFF
--- a/configs/sst_cs_apps-python-crb-eln.yaml
+++ b/configs/sst_cs_apps-python-crb-eln.yaml
@@ -24,7 +24,6 @@ data:
     - python3-psutil-tests
     - py3c-devel
     - py3c-doc
-    - marshalparser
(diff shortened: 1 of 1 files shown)


yselkowitz (changes requested):
The packages required for the ReproduciblePackageBuilds change have yet to be added to a workload or as a dependency in ELN.

hroncok:
I was going to say they are required by redhat-rpm-macros, but they are not: https://src.fedoraproject.org/rpms/redhat-rpm-config/blob/rawhide/f/redhat-rpm-config.spec#_118

I will figure the plan out.

PR #1194, opened 2024-06-22
sst_cs_system_management-text-processing: add missing texlive-lua-uni…

ngothan:
sst_cs_system_management-text-processing: add missing texlive-lua-uni-algos which is needed in texlive-luaotfload


yselkowitz (approved):
Filed https://src.fedoraproject.org/rpms/texlive/pull-request/11 for ELN.

PR #1197, opened 2024-06-25
Drop c9s

yselkowitz:
/cc @asamalik @sgallagher


yselkowitz:
> Something to consider: should we also replace the ELN view with "CentOS Stream 11" at this point? We'd need to change the config for ELNBuildSync slightly, but it might be more clear in the interface what we are tracking for each release.

ELN != CentOS Stream 11, they have a few different packages (fedora-* vs centos-*), built on different infrastructure, etc.  I think it would be confusing to call ELN c11s now and then totally change the source of c11s at branching.

PR #1198, opened 2024-06-27
Drop GNOME Logs from RHEL 10+

tpopela:
Blocks https://issues.redhat.com/browse/RHELMISC-5823


yselkowitz:
Have the anaconda wayland changes also landed in rawhide/ELN?

tpopela:
I don't think so, but I don't want to block the RHEL removal on it. So
should we just leave this open until the removal happens in Fedora and then
ELN and just proceed with the RHEL side of things? OTOH this will make the
workloads to error on c10s, unless I create a eln c10s specific version of
them and remove tigervnc from them. WDYT?

Dne čt 27. 6. 2024 15:41 uživatel Yaakov Selkowitz ***@***.***>
napsal:

yselkowitz:
The gnome-logs removal could go in now if you want to split that out, but we still need tigervnc in ELN until f41 anaconda implements the removal there.  The RHEL removal may proceed, and the c10s workloads will show warnings but not errors (the difference being that the rest of the packages in the workload will still be counted), but we will both know why so it shouldn't be an issue.

yselkowitz (approved):

PR #1199, opened 2024-06-27
Bring pcs dependenies up to date

CtrlZmaster:
This PR contains the up-to-date 1:1 mapping to Requires and BuildRequires in pcs spec file. I noticed that nobody updated the dependencies here since pcs became a placeholder. So I came to the conclusion that I, as a maintainer, should keep them up to date also here. I don't know how many placeholders exist, but I think that it would be nice to remind other maintainers that they should update their dependencies too. 

I switched to requiring virtual provides for pkgconfig in pcs spec file where possible. I suppose that Content Resolver doesn't support resolving that, so I left those dependencies as they were before.

I noticed that some dependencies were missing completely, like python or ruby for the srpm. Also, gcc and gcc-c++ were missing, these are needed for compiling python module and rubygem extensions. Maybe this was intentional, feel free to remove them, but I wanted to map the dependencies 1:1 to the spec file.

### DIFF
--- a/configs/sst_high_availability-userspace.yaml
+++ b/configs/sst_high_availability-userspace.yaml
@@ -164,27 +164,34 @@ data:
             - x86_64
     - srpm_name: pcs
       build_dependencies:
+        - autoconf
+        - automake
         - booth
+        - coreutils
         - corosync-qdevice-devel
         - corosynclib-devel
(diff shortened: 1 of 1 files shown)


yselkowitz (approved):
Thanks!

PR #1200, opened 2024-06-28
lorax-templates: Update the list for lorax-templates-rhel-10.0-49

bcl:
This following packages are no longer in the runtime-install.tmpl:

    dbus-x11
    gdisk
    tigervnc-server-minimal
    tigervnc-server-module
    xorg-x11-xauth
    xorg-x11-drivers
    xorg-x11-server-Xorg

Related: RHEL-38741

### DIFF
--- a/configs/sst_program-lorax-template-x11.yaml
+++ b/configs/sst_program-lorax-template-x11.yaml
@@ -1,0 +1,25 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: lorax template X11 packages
+  description: X11 Packages for ELN lorax template until switched to Wayland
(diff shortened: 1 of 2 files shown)


yselkowitz:
Have the corresponding anaconda changes reached rawhide yet?

yselkowitz:
Added an ELN config to carry the X11 packages in the meantime so this doesn't break.  Included the gdisk removal in https://src.fedoraproject.org/rpms/lorax-templates-rhel/pull-request/10 but other things depend on it so it won't disappear prematurely.

nielsdg:
Thanks both of you!

PR #1201, opened 2024-06-29
Drop docbook5-schemas

Lupphes:
Due to a licensing issue, we are dropping the docbook5-schemas support from RHEL. This is a test to see if that is a possibility or if it is still a hard requirement.

Related: RHELMISC-5885, BZ#2260534

### DIFF
--- a/configs/sst_cs_system_management-text-processing.yaml
+++ b/configs/sst_cs_system_management-text-processing.yaml
@@ -9,7 +9,6 @@ data:
     - asciidoc-doc
     - docbook-style-dsssl
     - docbook-utils
-    - docbook5-schemas
     - doxygen
     - doxygen-latex
     - help2man
--- a/configs/sst_cs_system_management-unwanted.yaml
+++ b/configs/sst_cs_system_management-unwanted.yaml
@@ -69,6 +69,7 @@ data:
     - bacula-client
     - bacula-console
     - udftools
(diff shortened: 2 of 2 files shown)


yselkowitz:
No problem unlisting it, but it won't be dropped from [the Content Resolver results](https://tiny.distro.builders/view-rpm--view-eln--docbook5-schemas.html) -- and hence can't be dropped from RHEL 10 -- until nothing depends on it anymore.  Right now that is pam and xmlstarlet, and you already have a PR open for the latter.

Do you also want to add it to https://github.com/minimization/content-resolver-input/blob/main/configs/sst_cs_system_management-unwanted.yaml to make it clear that it is an unwanted dependency?

Lupphes:
Thank you for your message! Yes, I'm currently removing both packages from the RHEL10. `xmlstarlet` should be deprecated soon, and  `pam` should hopefully be.

I wonder if marking the package as unwanted is necessary, as I don't know what it implies. However, I can definitely add it there.

yselkowitz:
Also adding it to unwanted would be desired here, as it would flag if something adds it as a dependency in the future.

Lupphes:
Ah, amazing. Alright will do with the next commit

yselkowitz:
Patch needs to be rebased.

Lupphes:
All rebased.

PR #1206, opened 2024-07-03
Convert all placeholders to new syntax, add openssl-fips-provider

yselkowitz:
/cc @asamalik @tdawson


tdawson:
If Content Resolver can handle it, these look good.  I much prefer the binary rpms being a sub of source rpms instead of the other way around.

yselkowitz:
Right now, we have some placeholders using the new syntax and some using the old.  This converts the remaining uses of the old, which will then allow us to drop support for the old syntax from the code (~35 lines).

PR #1208, opened 2024-07-05
Update rng and qat subpackages for sst_kernel_security

nefigtut:


configs/sst_kernel_security-intel-qat.yaml
```diff
@@ -8,34 +8,16 @@ data:
...
+      - qatlib-service
+      - qatlib-tests
       - qatengine
       - qatzip
       - qatzip-devel
+      - qatzip-libs
+      - xxhash
+      - xxhash-devel
+      - xxhash-doc
+      - xxhash-libs
```
yselkowitz, line 20: Currently, xxhash is showing as a dependency of dwz and rsync.  What is the relation to QAT and x86_64 specifically?

yselkowitz (commented):
Thanks for the PR!  One question for you...

configs/sst_kernel_security-intel-qat.yaml
nefigtut: Hi, Yaakov, there is a story behind this. Intel has requested us to get xxhash into RHEL as a future dependency for qatzip. So I guess it was before dwz and rsync. It is no there yet, but last time I've clarified, Intel confirmed they are refactoring qatzip to use xxhash. This is why we need xxhash and it should be a part of the QAT pack.

PR #1212, opened 2024-07-12
Replace gnome-terminal with ptyxis

yselkowitz:
/cc @tpopela


tpopela:
Thank you Yaakov!

Dne pá 12. 7. 2024 23:20 uživatel Yaakov Selkowitz ***@***.***>
napsal:

PR #1216, opened 2024-07-16
Update nodejs workload

khardix:
In reaction of packaging changes in Fedora (soon to be in RHEL).

- Clarify this is for the default NodeJS stream; other streams' packages will be named nodejsXY-<pkg>
- Rename npm to nodejs-npm
- Fix typo

---

I also have a few questions I would like to clarify:

1. For the future alternative streams (i.e. nodejs22, nodejs24, …), should new workload be created? These should be independent of one another (parallel installable), so I'm assuming that they should.
2. The "default" packages can be generated from different SRPMs based on which macros are set. (I.e. `nodejs20` SRPM currently produces `nodejs`, `nodejs-npm`, … packages, but can suddently start producing `nodejs20`, `nodejs20-npm`, …) Is this something I should somehow inform content resolver about, or it does not care and simply deals with binary rpm names no matter which SRPM they come from?
3. Can I submit workload for "nonexistent" RPMs? (In preparation for the name change described above, for example.)

### DIFF
--- a/configs/sst_cs_apps-nodejs.yaml
+++ b/configs/sst_cs_apps-nodejs.yaml
@@ -1,13 +1,13 @@
 document: feedback-pipeline-workload
 version: 1
 data:
-  name: Node.js runtime
-  description: Node.js (Javascript) runtime and NPM installer for appliechoion development
+  name: Node.js runtime – default stream
+  description: Node.js (Javascript) runtime and NPM installer for application development
   maintainer: sst_cs_apps
   packages:
     - nodejs
     - nodejs-devel
-    - npm
+    - nodejs-npm
     - nodejs-nodemon
   labels:
     - eln


yselkowitz:
>     1. For the future alternative streams (i.e. nodejs22, nodejs24, …), should new workload be created? These should be independent of one another (parallel installable), so I'm assuming that they should.

Yes, but a single file should suffice for all of them.  Also, that new workload for non-default streams should be tagged `c10s`-only, not `eln`, as ELN will continue to follow rawhide in preparation for RHEL 11.
 
>     2. The "default" packages can be generated from different SRPMs based on which macros are set. (I.e. `nodejs20` SRPM currently produces `nodejs`, `nodejs-npm`, … packages, but can suddently start producing `nodejs20`, `nodejs20-npm`, …) Is this something I should somehow inform content resolver about, or it does not care and simply deals with binary rpm names no matter which SRPM they come from?

The latter, CR will figure this out on its own.

>     3. Can I submit workload for "nonexistent" RPMs? (In preparation for the name change described above, for example.)

Doing so would cause a warning, so best to leave it until you're ready to add them.

yselkowitz (approved):

PR #1220, opened 2024-07-23
Remove qt6-qtbase-private-devel as it should be shipped in CRB

tpopela:
And being specified here makes it go into AppStream


tpopela:
Ah! Thank you @yselkowitz !

PR #1223, opened 2024-07-26
Add libblockdev-smart subpackages

tbzatek:
Planned libblockdev and udisks2 rebase.

Cc: @vojtechtrefny @dwlehman

### DIFF
--- a/configs/sst_storage_management-packages.yaml
+++ b/configs/sst_storage_management-packages.yaml
@@ -36,6 +36,8 @@ data:
     - libblockdev-nvme
     - libblockdev-part
     - libblockdev-plugins-all
+    - libblockdev-smart
+    - libblockdev-smartmontools
     - libblockdev-swap
     - libblockdev-tools
     - libblockdev-utils


yselkowitz:
@tbzatek Fedora doesn't have these subpackages yet.  Also, please note that libblockdev was F41FTBFS.

vojtechtrefny:
> @tbzatek Fedora doesn't have these subpackages yet. 

The new plugin will be available in Fedora after the next upstream release  (which will be hopefully in about a month).

> Also, please note that libblockdev was F41FTBFS.

That's fixed now. (I didn't get any notification about the failed mass rebuild.)

tbzatek:
> @tbzatek Fedora doesn't have these subpackages yet.

Yes, F41 development deadline is later than RHEL 10.0 beta deadline.

This is currently blocking gating on c10s due to the subpackages being filtered out in `osci.brew-build.rpmdeplint.functional` tests.

yselkowitz:
While we do want to get this into CR for tracking purposes, CR does not currently define the content of composes.  That test failure appears to be related to these subpackages being missing from the composes themselves, which makes sense given that they are new.  I have added this to my current pungi-centos MR:

https://gitlab.com/redhat/centos-stream/release-engineering/pungi-centos/-/merge_requests/716

These subpackages will show up in CR for c10s once they are in a compose by virtue of the fact that they are dependencies of -plugins-all, albeit as a Dependency rather than the intended Required.  But if I merge this now, it will trigger warnings in ELN which (following rawhide) doesn't have these yet.  So I'd prefer to leave this open until then.

tbzatek:
Thanks, we'll waive the gating test for the moment to push it through.

Theoretically, the `libblockdev-smart` subpackage could stay marked just as a 'dependency' and `udisks2` + `libblockdev-plugins-all` will pull it in. However we've explicitly listed other required plugins already, so staying consistent.

tbzatek:
Checked `RHEL-10.0-20240819.1`, both the `libblockdev-smart-3.1.0-7.el10` + `libblockdev-smartmontools-3.1.0-7.el10` and `udisks2-2.10.90-2.gitdb54112e.el10` are there. I've also added records to PES.

yselkowitz (approved):
These have now landed in ELN.

PR #1224, opened 2024-07-26
Drop redis from ELN

yselkowitz:
redis has been retired from F41:  https://fedoraproject.org/wiki/Changes/Replace_Redis_With_Valkey

/cc @notroj

### DIFF
--- a/configs/sst_cs_infra_services-redis.yaml
+++ b/configs/sst_cs_infra_services-redis.yaml
@@ -9,5 +9,4 @@ data:
     - redis-doc
     - redis-devel
   labels:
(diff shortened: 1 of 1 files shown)


notroj:
Thanks for the CC. +1

Conan-Kudo:
Shouldn't this have been renamed and updated for Valkey instead?

yselkowitz:
> Shouldn't this have been renamed and updated for Valkey instead?

That was #1154

Conan-Kudo:
Ah, thanks for that. 👍🏾

PR #1230, opened 2024-07-31
Drop redis from c10s

yselkowitz:
One question, is `valkey-compat(-redis)` going to be shipped?  If so, can the name be finalized?

/cc @notroj


notroj:
Currently no we're not planning to ship `valkey-compat` though it is under consideration. https://issues.redhat.com/browse/RHEL-51457

PR #1231, opened 2024-08-01
Drop babl

yselkowitz:


jridky:
Confirmed. Babl should be dropped.

PR #1237, opened 2024-08-02
Fix arches of kaidan and plasma-nm-l2tp

yselkowitz:
kaidan depends on qtwebengine.  plasma-nm-l2tp transitively depends on xl2tpd which requires kmod(l2tp_ppp.ko), which is not included in the RHEL kernel for s390x.

/cc @tdawson


tdawson:
LGTM
Thank You

PR #1243, opened 2024-08-06
Add automotive packages to ELN extras

sandrobonazzola:
Adding Automotive SIG packages to ELN Extras.

### DIFF
--- a/configs/eln_extras_automotive.yaml
+++ b/configs/eln_extras_automotive.yaml
@@ -1,0 +1,22 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Automotive packages
+  description: ELN Extras packages that the CentOS Automotive SIG cares about
+  maintainer: sig-automotive
+  packages:
+    - bluechi
+    - qm
+  arch_packages:
+    x86_64:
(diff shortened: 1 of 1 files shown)


configs/eln_extras_automotive.yaml
```diff
@@ -0,0 +1,15 @@
...
+version: 1
+data:
+  name: Automotive packages
+  description: ELN Extras packages that the CentOS Automotive SIG cares about
+  maintainer: sig-automotive
+  packages:
+    - qm
+  arch_packages:
+    x86_64:
+      - vhost-device-*
```
sandrobonazzola, line 11: vhost-device-* packages are built in koji for rust-vhost-device* src.rpm. Is it correct here listing the built binaries rpms? Or should I add here the package name as defined in koji/src.rpm?

yselkowitz: RPM name is correct for `packages` and `arch_packages`.

sandrobonazzola:
> The asterisks need to be replaced with complete package (RPM) name(s).

Done

yselkowitz (approved):

PR #1246, opened 2024-08-06
Snapshot, Loupe and Brasero are not shipped on s390x/aarch64

tpopela:
Reflect that status quo from RHEL 8+ here as well


yselkowitz:
I have to ask the obvious question, *why* are these shipped only on P&X?  At least wrt Snapshot and Loupe, I can't think of a technical reason.

tpopela:
webcams and image viewing are not supported use cases on servers. It's not about technical reasons, but about supportability.

PR #1248, opened 2024-08-06
Remove iperf3-devel.

notroj:


yselkowitz (approved):
Nothing should need this.

PR #1249, opened 2024-08-07
Don't include libstoragemgmt-smis-plugin or python3-pywbem in Storage…

dwlehman:
… Management


yselkowitz (approved):
OK, but pywbem is also a builddep of libstoragemgmt, so you'll need something like https://src.fedoraproject.org/rpms/libstoragemgmt/pull-request/6 to fully remove it.

PR #1254, opened 2024-08-08
Mark totem-pl-parser as unwanted

yselkowitz:
/cc @tpopela


tpopela:
The totem-pl-parser is unwanted by sst-desktop-platform-technologies, so it should be moved there.

yselkowitz:
Modified.

tpopela:
Great, thank you @yselkowitz !

PR #1259, opened 2024-08-12
Add rust debugger and wasm32 subpackages

yselkowitz:
/cc @cuviper

### DIFF
--- a/configs/sst_pt_llvm_rust_go-rust-toolset.yaml
+++ b/configs/sst_pt_llvm_rust_go-rust-toolset.yaml
@@ -10,7 +10,12 @@ data:
     - clippy
     - rust-analyzer
     - rust-doc
+    - rust-gdb
+    - rust-lldb
     - rust-std-static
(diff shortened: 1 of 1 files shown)


configs/sst_pt_llvm_rust_go-rust-toolset.yaml
```diff
@@ -10,7 +10,12 @@ data:
     - clippy
     - rust-analyzer
     - rust-doc
+    - rust-gdb
+    - rust-lldb
     - rust-std-static
+    - rust-std-static-wasm32-unknown-unknown
+    - rust-std-static-wasm32-wasi
```
cuviper, line 17: Note that plain `wasi` will be removed in favor of `wasip1` in the near future:
https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html#renaming-wasm32-wasi-to-wasm32-wasip1

But for now, it's correct to include both.

PR #1261, opened 2024-08-13
Drop xxhash package from RHEL10

nefigtut:
xxhash is not required by any of QAT packages.
dwz and rsync have dropped xxhash requirement too.


yselkowitz:
Both packages still have build dependencies on xxhash-devel that need to be fixed before xxhash can be fully removed, but it's okay to delist them as explicitly wanted in the meantime.

nefigtut:
thanks, Yaakov @yselkowitz, your help is most appreciated.

PR #1272, opened 2024-08-21
sst_gpu-X11-environment: Add xrdb

nielsdg:
It makes sense to ship this while we still have support for X11 clients, and is actively used by other packages that we still ship for configuration of their drawing (Xt, Xaw, etc).

Related: https://issues.redhat.com/browse/GPU-1145


nielsdg:
jfyi, @bstinsonmhk

yselkowitz:
Would you like to keep this alphabetical?

yselkowitz (approved):

PR #1273, opened 2024-08-21
ssg_idm-unwanted: Remove python3-netifaces

rjeffman:
The upstream project is no longer maintained and the package will be replaced by python3-ifaddr.

### DIFF
--- a/configs/ssg_idm-unwanted.yaml
+++ b/configs/ssg_idm-unwanted.yaml
@@ -5,6 +5,8 @@ data:
   description: Packages that the Identity Management subsystem does not want to distribute in RHEL 9
   maintainer: ssg_idm
   unwanted_packages:
+    # Removed in RHEL 10
+    - python3-netifaces
     # Drop support from RHEL-9 (replaced by SSSD)
     - nss-pam-ldapd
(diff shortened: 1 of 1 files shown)


configs/ssg_idm-unwanted.yaml
```diff
@@ -5,6 +5,8 @@ data:
   description: Packages that the Identity Management subsystem does not want to distribute in RHEL 9
   maintainer: ssg_idm
   unwanted_packages:
+    # Removed in RHEL 10
+    - python-netifaces
```
yselkowitz, line 9: `unwanted_packages` are RPMs, where `unwanted_source_packages` are SRPMs.  While the latter is preferred, in the meantime this needs to be:
```suggestion
    - python3-netifaces
```

yselkowitz (changes requested):

configs/ssg_idm-unwanted.yaml
rjeffman: Fixed. Thank you for the info.

yselkowitz (approved):
At risk of stating the obvious, you'll still need to update python3-ipalib in c10s to use ifaddr instead of netifaces; ELN looks taken care of already.

PR #1275, opened 2024-08-22
Drop google-benchmark

liutgnu:
Ticket: RHELMISC-6531

### DIFF
--- a/configs/sst_kernel_debug-unwanted.yaml
+++ b/configs/sst_kernel_debug-unwanted.yaml
@@ -8,6 +8,7 @@ data:
     - netdump-server
     - crash-ptdump-command
     - python-dmidecode
+    - google-benchmark
   labels:
     - eln
     - c10s


yselkowitz (changes requested):
That won't quite do what you're intending.  Instead, I suggest changing `unwanted_packages` to `unwanted_source_packages` and dropping the defunct `netdump-server` and `crash-ptdump-command`.

liutgnu:
Hi @yselkowitz ,

> That won't quite do what you're intending. Instead, I suggest changing `unwanted_packages` to `unwanted_source_packages` and dropping the defunct `crash-ptdump-command`.

Thanks for your help. I guess the MR is no longer needed because the package has already been dropped and the process of dropping has been done. So I will close the MR. Thanks again for your help!

Thanks,
Tao Liu

yselkowitz:
Marking a package as unwanted never triggered its removal, but the PR is still useful as it will flag this should another package begin to depend on this in the future.

PR #1279, opened 2024-09-04
Move x3270 package under arch_hw as owners

jkonecny12:
The x3270 is not maintained nor direct dependency of sst_installer. Not sure why we own it. I contacted jcajka about this and he agreed to move this package under arch_hw to make this consistent with bugzilla-data.


jcajka (approved):
LGTM

PR #1280, opened 2024-09-04
kde: marknote replaces knotes

yselkowitz:
/cc @tdawson

### DIFF
--- a/configs/eln_extras_fedora-packager.yaml
+++ b/configs/eln_extras_fedora-packager.yaml
@@ -1,0 +1,18 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Fedora Packager
+  description: Tools and utilities needed by a Fedora Packager
+  maintainer: epel-sig
+  packages:
+    - bodhi-client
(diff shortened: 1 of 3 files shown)


sgallagher (approved):
The Fedora packager commit looks good to me.

tdawson:
The KDE stuff looks good.
Why do we have git in Fedora Packager, and eln-extra group?

tdawson:
Actually, let me take back my statement about KDE.  Why are we replacing knote with marknote.  From what I see, they do two totally different things.

yselkowitz:
see: https://invent.kde.org/pim/pim-technical-roadmap/-/issues/58

yselkowitz:
see https://invent.kde.org/pim/pim-technical-roadmap/-/issues/58

tdawson:
Thank you for the explanation.  If that is the case, then I give a thumbs up to the KDE section.

yselkowitz:
> Why do we have git in Fedora Packager, and eln-extra group?
 Copied straight from the epel9 comps.

PR #1290, opened 2024-09-30
Add deps of nodejs for proper WASM rebuilds

hhorak:


yselkowitz:
These are still bundled in rawhide and hence ELN.  Will that be changing?

hhorak:
> These are still bundled in rawhide and hence ELN. Will that be changing?

Yes, this should get unbundled in Fedora and ELN as well -- @khardix are there some unresolved obstacles for the unbundling in Fedora Rawhide that we track?

yselkowitz:
First off, nodejs-cjs-module-lexer is currently FTBFS: https://koji.fedoraproject.org/koji/packageinfo?packageID=38748

yselkowitz:
And a attempted scratch rebuild of wasi-libc also failed: https://koji.fedoraproject.org/koji/taskinfo?taskID=124251326

khardix:
> @khardix are there some unresolved obstacles for the unbundling in Fedora Rawhide that we track?

Still the same – how to make the re-built modules shared between streams. I did not find an existing ticket, so I filed one: https://bugzilla.redhat.com/show_bug.cgi?id=2316095

> nodejs-cjs-module-lexer is currently FTBFS

Due to the general nodejs propensity for bundling and adding new (build) dependencies whenever a shiny new toy appear, intermittent FTBFS are to be expected, unfortunately. I'm aware of the issue and working on a solution.

> wasi-libc also failed

There is a test that checks that the macros defined in various LIBC headers are as expected; it needs to be updated manually whenever the build environment changes. I'll take look at it.

yselkowitz (approved):
All these build again successfully in ELN.

PR #1294, opened 2024-10-08
sst_csi_client_tools: add yara

mhuth:
This RHEL-only package exists already in 8 and 9, and it was forgotten from 10

### DIFF
--- a/configs/sst_csi_client_tools.yaml
+++ b/configs/sst_csi_client_tools.yaml
@@ -14,6 +14,7 @@ data:
     - virt-who
     - yggdrasil
     - yggdrasil-worker-package-manager
+    - yara
   package_placeholders:
     - srpm_name: insights-client
       rpms:


yselkowitz:
Why exactly is this RHEL specific, it exists in Fedora too: https://src.fedoraproject.org/rpms/yara/blob/rawhide/f/yara.spec

mhuth:
@yselkowitz Because we are adding it to RHEL10?  I was just following the instructions from here: https://one.redhat.com/rhel-development-guide/#proc_adding-new-packages_assembly_managing-RHEL-content
... and it says to add it to the content-resolver, so here I am!

yselkowitz (changes requested):
A RHEL-specific package means one that does not exist in Fedora, therefore cannot be built in ELN and a placeholder is used instead.  yara is a Fedora package though and therefore is not RHEL-specific.  Therefore, `yara` should simply be added to `.data.packages[]` as a regular package.

yselkowitz (approved):

yselkowitz:
This doesn't mean that you won't build and ship the code in RHEL, but it does mean that we track it as a regular package, and start building it in ELN in preparation for RHEL 11.

yselkowitz:
Please note that an unwanted dependency was found: https://src.fedoraproject.org/rpms/yara/pull-request/4

mhuth:
@yselkowitz I had to add this to the yara spec when building it for RHEL10 (compared to RHEL8 & 9):
```
%if 0%{?rhel} && 0%{?rhel} <= 9
BuildRequires:  pcre
%else
BuildRequires:  pcre2
%endif
```

yselkowitz:
AFAICS pcre is not used at all in the build.

mhuth:
You are correct.  It wasn't needed afterall /shrug

PR #1295, opened 2024-10-08
Initial ELN Extras Python config

hroncok:
The packages are unwanted in eln, but desired in eln extras. I hope that's alright.


yselkowitz (approved):
Yes, that should be doable now.

PR #1301, opened 2024-10-21
Add dotnet9.0

yselkowitz:
/cc @omajid

### DIFF
--- a/configs/sst_dotnet-dotnet.yaml
+++ b/configs/sst_dotnet-dotnet.yaml
@@ -9,6 +9,10 @@ data:
     - dotnet-sdk-8.0
     - dotnet-runtime-8.0
     - aspnetcore-runtime-8.0
+    # EOL in May, 2026
+    - dotnet-sdk-9.0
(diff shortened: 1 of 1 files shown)


omajid:
Please note that the EOL of dotnet8.0 is later. Should we add dotnet9.0? This package is basically the same scenario as dotnet7.0 which was put in unwanted instead: https://github.com/minimization/content-resolver-input/pull/1040

yselkowitz:
7.0 was marked unwanted when 8.0 was already around to supersede it.  As it looks like 9.0 is going to be in c10s, we want to track it (at least there) and not mark it unwanted yet.  Once 10.0 is out (late next year?) then we'll want to mark 8.0 and 9.0 unwanted in ELN, and drop them from c10s once they go out of support.

Another advantage of tracking 9.0 in ELN is the potential to find issues leading up to 10.0 which we wouldn't necessarily get by only tracking 8.0.

omajid:
Okay, that works for me!

PR #1302, opened 2024-10-21
Remove yajl

yselkowitz:
libvirt has switched to json-c, and crun uses an embedded yajl as of RHEL 10.

https://src.fedoraproject.org/rpms/crun/pull-request/72
https://github.com/containers/crun/pull/1583
https://gitlab.com/redhat/centos-stream/rpms/crun/-/merge_requests/102

/cc @jnovy @cgwalters


jnovy:
LGTM, thanks @yselkowitz !

PR #1303, opened 2024-10-22
add new subpackage opencryptoki-ccatok for rhel-10.0/c10s

ngothan:
add new subpackage opencryptoki-ccatok for rhel-10.0/c10s

### DIFF
--- a/configs/sst_cs_base_utils-sys-management.yaml
+++ b/configs/sst_cs_base_utils-sys-management.yaml
@@ -79,6 +79,7 @@ data:
       - setserial
     x86_64:
       - setserial
+      - opencryptoki-ccatok
     ppc64le:
       - iprutils
       - setserial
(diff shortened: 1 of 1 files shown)


ngothan:
Hi @yselkowitz, could you please help me to merge this. It's needed for https://issues.redhat.com/browse/RHEL-39004

https://errata.devel.redhat.com/advisory/140177/builds

Thank you!

yselkowitz (changes requested):
First off, this needs to be rebased against `main`; this should only consist of a single commit.

Furthermore, `opencryptoki-ccatok` is not built for aarch64, therefore it cannot be listed unconditionally.  Instead, the existing s390x listing must remain, and this package be added to the ppc64le and x86_64 lists.

Finally, while we do want this fixed here, Content Resolver does not determine the content of composes.  Therefore, this is not the blocking factor for your errata.  In order to add this subpackage to other arches, you will need to file an ENGCMP ticket to have it added on those arches in pungi-centos.

ngothan:
@yselkowitz , i rebased the PR, keeped the s390x listing and added opencryptoki-ccatok in x86_64 and ppc64le list

yselkowitz (approved):

PR #1304, opened 2024-10-23
Drop obsolete texlive-ms

yselkowitz:
This was replaced with separate count1to and multitoc subpackages in texlive 2024.

/cc @ngothan


yselkowitz:
ping @ngothan

ngothan:
Hi @yselkowitz , texlive-ms is splitted into multitoc and count1to in texlive-2024. I  forgot to remove it from content-resolver-input. Thanks for the fix!

PR #1305, opened 2024-10-24
Finish component transfer to `sst_cs_base_utils`

lzaoral:


configs/sst_cs_base_utils-sys-management.yaml
```diff
@@ -5,17 +5,25 @@ data:
   description: Packages for managing Linux system
   maintainer: sst_cs_base_utils
   packages:
+    - acl
+    - attr
     - basesystem
     - bc
+    - coreutils
+    - coreutils-single
```
yselkowitz, line 13: CR installs each workload as a group.  Since coreutils and coreutils-single conflict with each other, they cannot be in the same workload, otherwise an error will result.

yselkowitz (changes requested):

yselkowitz (approved):

PR #1306, opened 2024-10-29
KDE: disable apps still based on KF5

yselkowitz:
EPEL 10 will include parts of Qt5 but not KF5.  Some of these will be ported to KF6 in KDE Gear 24.12, at which point they can be re-enabled.

/cc @tdawson

### DIFF
--- a/configs/eln_extras_kde.yaml
+++ b/configs/eln_extras_kde.yaml
@@ -6,12 +6,12 @@ data:
   maintainer: tdawson
   packages:
     - alligator
-    - amarok
+    #- amarok
     - analitza
     - ark
     - ark-libs
-    - artikulate
-    - artikulate-libs
+    #- artikulate
+    #- artikulate-libs
     - atlantik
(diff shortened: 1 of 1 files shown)


yselkowitz:
@tdawson FYI that removed most of KF5, except for kwayland and kwindowsystem as dependencies of kwayland-integration, which (while Qt5-only) is still maintained as part of Plasma and is listed as a hard dependency of plasma-workspace-wayland.  Does that still make sense for a KF6-only EPEL 10?

tdawson:
You are correct, we will need to conditionalize kwayland-integration out of plasma-workspace-wayland.
I haven't touched plasma-workspace yet due to it's build dependencies not being met yet.
I should have time to do a merge request to fix  it tomorrow.

tdawson:
I guess I had time today.  Here is the merge request
https://src.fedoraproject.org/rpms/plasma-workspace/pull-request/32

PR #1307, opened 2024-10-29
Add gvisor-tap-vsock

yselkowitz:
This replaces libslirp in the containers stack as of RHEL 9.

/cc @jnovy @cgwalters


jnovy:
LGTM, thanks.

PR #1313, opened 2024-11-12
papers replaces evince in RHEL 10

yselkowitz:
/cc @tpopela


tpopela:
Thank you @yselkowitz !

PR #1314, opened 2024-11-13
Add php-pecl-redis6 to the PHP workload.

notroj:


yselkowitz (changes requested):
@notroj it looks like this will result in the addition of liblzf, php-pecl-igbinary, and php-pecl-msgpack (although the last two are already conditionalized and could be disabled if unwanted) as dependencies.  Could you please confirm what the plans are for those?

Also, do you want to alphabetize this (at least within the `php-pecl-*` group), and take the opportunity to clean up the duplicate `php-pear` and `php-pecl-zip` too?

notroj:
We're going to flip the bconds to default-off for RHEL

notroj:
> Also, do you want to alphabetize this (at least within the `php-pecl-*` group), and take the opportunity to clean up the duplicate `php-pear` and `php-pecl-zip` too?

Thanks hadn't spotted it - I will be updating #1315 to move these components to a new SST, I'll do it then.

yselkowitz:
And liblzf will end up added, or are you going to use the bundled version for RHEL/ELN?

yselkowitz:
@notroj https://src.fedoraproject.org/rpms/php-pecl-redis6/pull-request/1

yselkowitz (approved):
The package has been conditionalized for ELN.

PR #1315, opened 2024-11-13
SST reorg: Infra Services split, merge with remainder of Apps.

notroj:
WIP, more to come.


hhorak:
LGTM, `sst_cs_apps-swig.yaml` can be moved together with those.

yselkowitz:
@notroj is this ready now?

configs/sst_cs_net_perf_services-terminal-emulator.yaml
```diff
@@ -3,11 +3,8 @@ version: 1
 data:
   name: Terminal Emulator
   description: Terminal emulators and fullscreen terminal applications/libraries
-  maintainer: sst_cs_infra_services
+  maintainer: sst_cs_net_perf_services
   packages:
-    - dialog
-    - minicom
-    - xterm
```
yselkowitz, line 10: That leaves this config without any packages.  Either minicom/xterm need to be here (instead of sst_cs_net_perf_services-miscellany.yaml), or this file should be removed entirely.

configs/sst_cs_stacks-imap-server.yaml
```diff
@@ -6,8 +6,8 @@ data:
   maintainer: sst_cs_infra_services
   packages:
     - cyrus-imapd
-    - procmail
     - xapian-core
+    - imaptest
```
yselkowitz, line 10: This would be a net addition, as the previous config which listed it was labeled `eln-test` (which is invalid).  FWIW imaptest has not been shipped in any previous RHEL versions.  Do you actually plan on adding it?

yselkowitz (changes requested):

notroj:
Thanks for the detailed review @yselkowitz

yselkowitz (approved):

PR #1317, opened 2024-11-14
Add sst_rhel_lightspeed workload config

r0x0d:


configs/sst_rhel_lightspeed.yaml
```diff
@@ -0,0 +1,24 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: RHEL Lightspeed
+  description: Command Line Assistant powered by RHEL Lightspeed.
+  maintainer: sst_rhel_lightspeed
+  packages:
+    - command-line-asssitant
```
yselkowitz, line 8: `packages` and `package_placeholders` are mutually exclusive.  Since this is a RHEL-specific package, the placeholder is appropriate, and therefore:
```suggestion
  packages: []
```

configs/sst_rhel_lightspeed.yaml
```diff
@@ -0,0 +1,24 @@
...
+  package_placeholders:
+    - srpm_name: command-line-assistant
+      rpms:
+        - rpm_name: command-line-assistant
+          description: Command Line Assistant
+          dependencies:
+            - tomli
+  labels:
+    - c9s
+    - c10s
```
yselkowitz, line 18: c9s is no longer tracked in Content Resolver, but if you expect to have this in RHEL 11, then ELN should be tracked instead (even as a placeholder):
```suggestion
  labels:
    - eln
    - c10s
```

configs/sst_rhel_lightspeed.yaml
```diff
@@ -0,0 +1,24 @@
...
+  maintainer: sst_rhel_lightspeed
+  packages:
+    - command-line-asssitant
+  package_placeholders:
+    - srpm_name: command-line-assistant
+      rpms:
+        - rpm_name: command-line-assistant
+          description: Command Line Assistant
+          dependencies:
+            - tomli
```
yselkowitz, line 15: Based on your copr build, and that only RHEL 10+ is tracked in Content Resolver:
```suggestion
            - python3
```

configs/sst_rhel_lightspeed.yaml
```diff
@@ -0,0 +1,24 @@
...
+  labels:
+    - c9s
+    - c10s
+
+  arch_packages:
+    x86_64:
+    - command-line-assistant
+    aarch64:
+    - command-line-assistant
```
yselkowitz, line 24: `arch_packages` is not for placeholders; instead, use `limit_arches` in the placeholder definition..

yselkowitz (changes requested):

configs/sst_rhel_lightspeed.yaml
r0x0d: updated

r0x0d:
Updated the definitions, @yselkowitz! Could you do another review of this whenever you have time?

yselkowitz (approved):

PR #1318, opened 2024-11-15
Update for sst_display_desktop_foundation (old sst_desktop) and sst_display_productivity (old sst_desktop_applications) changes

tpopela:


yselkowitz (changes requested):
Thanks @tpopela.  There are still two workloads named `GNOME Desktop` with the same description `GNOME Desktop core packages`, albeit from two different SSTs.  Could these be differentiated somehow in their name and description?

tpopela:
We used to have this for a long time so I would prefer to leave it as it is
Yaakov. And in the end it's the same workload, but we're just moving
packages around.

Dne pá 15. 11. 2024 19:51 uživatel Yaakov Selkowitz <
***@***.***> napsal:

yselkowitz (approved):

PR #1319, opened 2024-11-19
Move components ownership from Apps to Base utils

hhorak:
@lzaoral FYI


configs/sst_cs_base_utils-unwanted.yaml
```diff
@@ -75,6 +75,7 @@ data:
     - python3-bottle
     - python3-pycurl
     - vorbis-tools
+    - mercurial # Hard to maintain while not the most popular choice anymore.
```
yselkowitz, line 78: Did you mean to also remove `mercurial` from https://github.com/minimization/content-resolver-input/blob/main/configs/sst_cs_apps-unwanted.yaml#L10 ?

yselkowitz (changes requested):

configs/sst_cs_base_utils-unwanted.yaml
hhorak: yes, it's now done as part of https://github.com/minimization/content-resolver-input/pull/1321

yselkowitz (approved):

PR #1326, opened 2024-11-25
Restore kmix, kwave, marble to Extras

yselkowitz:
These have migrated to KF6 in KDE Gear 24.12.

/cc @tdawson


tdawson:
LGTM
Thank you for that.

PR #1328, opened 2024-11-26
RHELPLAN-171066 - add xxhash to gcc-misc

fche:
cc: @hhorak


configs/sst_pt_gcc-misc.yaml
```diff
@@ -10,6 +10,10 @@ data:
     - binutils-gold
     - annobin-annocheck
     - annobin-plugin-gcc
+    # used by several sst_pt tools (libabigail, dwz, binutils, debugedit)
+    - xxhash
+    - xxhash-libs
+    - xxhash-devel
```
siddhesh, line 16: fa5645be2dd1ccd543e286e8137654d36e8b6377 also had `xxhash-docs`, but I reckon we don't need it if we keep it as buildroot-only.

siddhesh (commented):
I reckon the xxhash entry also needs to be dropped from `configs/sst_kernel_security-unwanted.yaml`.

configs/sst_pt_gcc-misc.yaml
yselkowitz: There's been some back-and-forth on xxhash; last I heard, it was going to be buildroot-only in order to build these tools package, and header-only mode so that there is no runtime dependency on xxhash-libs.  Is that still the case?  Do you plan on shipping any part of xxhash in CRB (technically not necessary)?

fche: I suspect we'd like to remain open to the possibility of CRB, not just buildroot-only, since some packages may want to use it in a mode not-only-inline.  Is there a downside in keeping the options open?

yselkowitz: In that case I think it would suffice to delist xxhash as unwanted but not add it here.  If it gets to the point that xxhash-libs is pulled in as a dependency (by a package using it *not* in inline mode) then it would make sense to ship the rest in CRB.

fche: OK, whatever works.  (I'm curious what the consequences are of listing the packages here, closer to their new packager sst, vs. not listing explicitly anywhere).

siddhesh: Lets start with delisting from unwanted and see where this goes.

fche: (By the way, how do we "see where this goes"?  The PRP docs talk about doing a "content resolver check", but don't describe how to check.)

siddhesh: I was going to say that it shouldn't show up [here](https://tiny.distro.builders/view-unwanted--view-c10s.html) but it already doesn't...

fche: https://github.com/minimization/content-resolver-input/pull/1329   is the "un-unwanted" alternative.

hhorak:
> cc: @hhorak

LGTM

configs/sst_pt_gcc-misc.yaml
yselkowitz: It won't show up in c10s CR until it is built again in c10s, so you can (and should) proceed with the PRP in the meantime.  Once built, xxhash(-devel) will show up again as a build dependency of at least dwz.

The question then remains if this will be buildroot only or will be shipped in CRB.

yselkowitz: This PR would be fine if/when you decide to add it to CRB.

yselkowitz:
Based on the discussion above, this package is going to be buildroot-only for now.  Therefore, I see no need to list it in CR, it will be (and in ELN already is) pulled in automatically as a build dependency of the affected packages (starting with dwz).  If at some point in the future you decide to ship xxhash -- e.g. if a package starts using it NOT in header-only inline mode, causing xxhash-libs to be a runtime dependency, and in which case the other subpackages should go into CRB -- then this would be the correct change to make.

PR #1331, opened 2024-11-28
Remove yara from sst_csi_client_tools

mhuth:


mhuth:
@ptoscano please do the needful.

ptoscano:
While removing it from sst_client_tools is correct, it will drop it entirely from the distro. yara **must** be in a workload to be shipped.

If yara belongs to sst_insights, then please move it to that workload.

mhuth:
@ptoscano It belongs to platmgmt-ins-security which isn't listed as a group in the config directory.  But it is a rover group: https://rover.redhat.com/groups/group/platmgmt-ins-security

ptoscano:
> It belongs to platmgmt-ins-security which isn't listed as a group in the config directory.

As I said, it must belong to a RHEL SST.

configs/sst_insights.yaml
```diff
@@ -4,7 +4,14 @@ data:
...
   maintainer: sst_insights
-  packages: []
+  packages:
+    - yara
+  package_placeholders:
+    - srpm_name: yara
+      rpms:
+        - rpm_name: yara
+          description: yara
+          dependencies: []
```
yselkowitz, line 14: See https://github.com/minimization/content-resolver-input/pull/1294#pullrequestreview-2353177631

yselkowitz (changes requested):

yselkowitz (approved):

PR #1332, opened 2024-11-28
Realign Content Resolver workloads post Dev/QE merger

nielsdg:
Some components got shuffled around, so it makes sense that we also align the content workloads with the new org structure.

For the unwanted packages, splitting this up over the multiple SSTs was a bit harder since they contains for example some very historical dependencies of e.g. Xorg or some multimedia libraries. I tried to move those the most appropriate SST where possible.

### DIFF
--- a/configs/sst_display_hardware_multimedia-unwanted.yaml
+++ b/configs/sst_display_hardware_multimedia-unwanted.yaml
@@ -11,6 +11,23 @@ data:
     - re2c
     # Totem is removed from RHEL 10, avoid usage in other components
     - totem-pl-parser
+    # Replaced by tools in libinput
+    - evemu
+    # BR for rubygem-clutter-gstreamer. Said rubygem shouldn't be in EL9. Also, superseeded by clutter-gst3.
+    - clutter-gst2
+    # Virtually unused even in Fedora, might have been used by spice once upon a time. Also, superseeded by opus.
+    - celt051
+    # Only used for some old games and emulators
+    - SFML
+    # Only used by some multimedia apps we don't ship
+    - libmpcdec
+    # Seems unused
+    - libsamplerate
+    # Only used by openal-soft examples, brings in other unwanted deps
+    - SDL_sound
+    # Removed in favor of pipewire(-pulse). Note that the SRPM with the same
+    # name should still be built for some packages, like pulseaudio-libs
+    - pulseaudio
   labels:
     - eln
     - c10s
--- a/configs/sst_gpu-unwanted.yaml
+++ b/configs/sst_gpu-unwanted.yaml
@@ -5,64 +5,11 @@ data:
   description: GPU - Packages which should not be in ELN
   maintainer: sst_gpu
   unwanted_packages:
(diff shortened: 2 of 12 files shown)


tpelka83:
LGTM to me.

Jiri-Koten:
LGTM

configs/sst_display_hardware_multimedia-unwanted.yaml
```diff
@@ -11,6 +11,23 @@ data:
...
+    # BR for rubygem-clutter-gstreamer. Said rubygem shouldn't be in EL9. Also, superseeded by clutter-gst3.
+    - clutter-gst2
+    # Virtually unused even in Fedora, might have been used by spice once upon a time. Also, superseeded by opus.
+    - celt051
+    # Only used for some old games and emulators
+    - SFML
+    # Only used by some multimedia apps we don't ship
+    - libmpcdec
+    # Seems unused
+    - libsamplerate
```
yselkowitz, line 25: Is this actually unwanted?  It is a dependency of alsa-plugins and alsa-utils, and has been in each RHEL release since at least 6.

configs/sst_display_hardware_multimedia-unwanted.yaml
```diff
@@ -11,6 +11,23 @@ data:
...
+    # Only used by openal-soft examples, brings in other unwanted deps
+    - SDL_sound
+    # Removed in favor of pipewire(-pulse). Note that the SRPM with the same
+    # name should still be built for some packages, like pulseaudio-libs
+    - pulseaudio
```
yselkowitz, line 30: FYI see https://issues.redhat.com/browse/RHELBU-537 and https://src.fedoraproject.org/rpms/pulseaudio/pull-request/12

configs/sst_gpu-unwanted.yaml
```diff
@@ -5,64 +5,11 @@ data:
   description: GPU - Packages which should not be in ELN
   maintainer: sst_gpu
```
yselkowitz, line 6: Is there still a `sst_gpu`?

nielsdg: Yep, and we're now (temporarily) our own new group: GPU & Accelerators .

configs/sst_display_hardware_multimedia-unwanted.yaml
nielsdg: Not sure, this just moves it from the unwanted sst_gpu packages to the sst_display_hardware_multimedia unwanted packages. It's been there for a pretty long time now, so I'll leave it to @Jiri-Koten  and @tpelka83 to figure that out. I don't have a strong opinion either way :-)

configs/sst_display_hardware_multimedia-unwanted.yaml
nielsdg: Yep, I know. It was already part of sst_gpu-unwanted.yml, so I guess this could be in a follow-up MR once those are resolved?

yselkowitz: This should definitely remain here unwanted, but even better if we don't build it all.

PR #1337, opened 2024-12-20
Skip rpmdistro-repoquery temporarily

yselkowitz:
This requires dnf4 which is no longer available for rawhide and ELN.


yselkowitz:
See https://pagure.io/rpmdistro-repoquery/pull-request/23 for background.

PR #1338, opened 2024-12-23
virtualization-all: add snphost and snpguest

yashkmankad:


yselkowitz (changes requested):
snphost is fine, but will eventually need vendoring for RHEL.  However, snpguest does not appear to be in Fedora (and hence not available for ELN)?

tylerfanelli:
> Hi @yselkowitz - yes, @tylerfanelli and I checked with Josh Stone, and they will be vendored in RHEL.
> 
> As to snpguest not being in Fedora yet, I think Tyler is waiting on a few dependencies to be built for Fedora before it can be added. eg: https://bugzilla.redhat.com/show_bug.cgi?id=2313237
> 

Just a note -- rust-x509-parser (linked) is the last dependency needed for snpguest. snpguest will be in Fedora very soon.

yselkowitz (approved):
(rust-)snpguest was added today to Fedora.

PR #1339, opened 2024-12-25
Reflect the sst split of sst-virt-firmware from sst-virtualization in Content Resolver

CongLi:


configs/sst_virtualization-firmware.yaml
```diff
@@ -0,0 +1,17 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Virtualization Firmware
+  description: Virtual firmware packages like EDK2, seabios which support virtualization capabilities
+  maintainer: sst_virt_firmware
```
yselkowitz, line 6: The convention is that filenames are prefixed with the SST name.  Therefore, this file should be named `sst_virt_firmware-SOMETHING.yaml`.

yselkowitz (changes requested):

CongLi:
@yselkowitz  Thanks for the explanation ! MR has been updated accordingly.

yselkowitz (approved):

PR #1342, opened 2025-01-02
meta: add poi-tracker

michel-slm:
Also manage the inventory with poi-tracker, which sorts the output

### DIFF
--- a/configs/eln_extras_meta.yaml
+++ b/configs/eln_extras_meta.yaml
@@ -22,24 +22,24 @@ data:
     - duo_unix-selinux
     - ebranch
     - et
-    - htop
-    - iotools
     - fd-find
     - fping
     - git-email
     - glslang
+    - htop
     - hyperfine
+    - iotools
     - keyrings-filesystem
     - lei
     - libdwarf
     - libdwarf-devel
-    - libkdumpfile-devel
     - libidn
     - libidn-devel
+    - libkdumpfile-devel
     - libxcrypt-compat
     - linux-sysinfo-snapshot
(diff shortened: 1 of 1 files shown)


configs/eln_extras_meta.yaml
```diff
@@ -54,20 +54,20 @@ data:
     - pam_krb5
     - parallel
     - pngquant
+    - poi-tracker
     - pore
     - pv
     - python3-pystemd
     - python3-testslide
     - python3-zstd
-#    - rpmdistro-repoquery
```
yselkowitz, line 62: This was disabled temporarily due to FTI; see https://pagure.io/rpmdistro-repoquery/pull-request/23

This should probably remain (albeit commented) until that is fixed.

yselkowitz (changes requested):

configs/eln_extras_meta.yaml
yselkowitz: rpmdistro-repoquery has been fixed and uncommented; this will need to be rebased but now that won't be lost either.

yselkowitz: > rpmdistro-repoquery has been fixed and uncommented; this will need to be rebased but now that won't be lost either.

ping @michel-slm

michel-slm: Ack, I need to add another package too. Will update after the weekend, last two weeks were PTO+buried in work

michel-slm: done

yselkowitz (approved):

PR #1344, opened 2025-01-08
virtualization-all: Add trustee-guest-components

kostyanf14:
Resolves: RHELPLAN-171373

### DIFF
--- a/configs/sst_virtualization-all.yaml
+++ b/configs/sst_virtualization-all.yaml
@@ -43,6 +43,7 @@ data:
       - spice-protocol
       - spice-vdagent
       - supermin
+      - trustee-guest-components
       - virt-p2v
       - virt-top
       - virt-v2v


configs/sst_virtualization-all.yaml
```diff
@@ -82,3 +82,16 @@ data:
           description: Add support for Windows guests to virt-v2v and virt-p2v
           dependencies:
             - libguestfs
+    # Package in RHEL and Fedora is different
+    - srpm_name: trustee-guest-components
+      build_dependencies:
+        - rust-toolset
+        - /usr/bin/pkg-config
```
yselkowitz, line 89: Let's not use file dependencies in Content Resolver; in this case, this is common enough that I think it can just be omitted here.

yselkowitz (changes requested):

configs/sst_virtualization-all.yaml
kostyanf14: fixed

configs/sst_virtualization-all.yaml
```diff
@@ -82,3 +82,15 @@ data:
           description: Add support for Windows guests to virt-v2v and virt-p2v
           dependencies:
             - libguestfs
+    # Package in RHEL and Fedora is different
```
yselkowitz, line 85: I don't see this package at all in Fedora; did you perhaps copy-and-paste this?

configs/sst_virtualization-all.yaml
```diff
@@ -82,3 +82,15 @@ data:
...
+      build_dependencies:
+        - rust-toolset
+        - openssl-devel
+        - tss2-devel
+        - tpm2-tss-devel
+      limit_arches:
+        - x86_64
+      rpms:
+        - rpm_name: trustee-guest-components
+          description: Tools that run in confidential VMs, attest and get secrets from Trustee
```
yselkowitz, line 96: Based on the build deps, I'd guess there will be some runtime deps too, along the lines of:
```suggestion
          description: Tools that run in confidential VMs, attest and get secrets from Trustee
          dependencies:
            - openssl-libs
            - tss2
            - tpm2-tss
```

yselkowitz (changes requested):

configs/sst_virtualization-all.yaml
kostyanf14: This package is currently under review for Fedora but it will be different. 
@uril please explain.

configs/sst_virtualization-all.yaml
kostyanf14: @uril Is this correct? I don't see runtime deps in the spec that you provided.

uril: I think all the runtime deps are automatically added when the package is built

configs/sst_virtualization-all.yaml
uril: In Fedora Rust packages that are dependencies are built as packages.
In RHEL, such packages are provided in a "vendor" tarball as a second source.

uril: Fedora Review Request bug: rhbz#2332550

uril: @yselkowitz Do you think we better list the package under "arch_packages: x86_64:" ?

configs/sst_virtualization-all.yaml
uril: You can manually add deps:
  openssl-libs
  tpm2-tss

configs/sst_virtualization-all.yaml
uril: Since the package is being added to Fedora too and the Fedora/RHEL difference is similar to other Rust packages.

configs/sst_virtualization-all.yaml
yselkowitz: Not to the spec file (as library dependencies are autogenerated), but they would be needed in the placeholder.  That being said, see my response to the other thread.

configs/sst_virtualization-all.yaml
yselkowitz: If the only difference between the Fedora and RHEL packages will be vendoring of rust dependencies, then don't use a placeholder.  An `eln` branch can be used to do the vendoring for ELN builds, or if nothing else, the vendoring will need to be done in c11s prior to branching.

kostyanf14: @yselkowitz Do you mean that we should add `trustee-guest-components` to `packages` and nothing else? 
The current situation is that we start RHEL PRP before the Fedora package is ready and we expect to have trustee-guest-components in RHEL 9.6/10 even if the Fedora package will be under review. 
How should we update the content-resolver in this case?

kostyanf14: @yselkowitz Do you have any suggestions?

uril:
> LGTM, will this be added to Fedora soon?

I hope it will. It got some preliminary reviews, but does not yet approved.
Review comments will be fixed in both Fedora and RHEL.

kostyanf14:
@yselkowitz When do you plan to merge this PR?

yselkowitz:
Adding a package to CR before it has been added to a distro just results in a warning message for a "missing" package, so I would normally wait until it has been added to either Fedora or RHEL; neither has happened yet.  You may continue with the PRP in the meantime.

kostyanf14:
@yselkowitz 

>You may continue with the PRP in the meantime.

Do I just mark RHELPLAN-171316 task as done and continue?

yselkowitz:
That's a RHEL9 PRP; CR only tracks c10s/RHEL10 and ELN.  You can mark that done, and when you get to that step in the RHEL10 PRP, this will cover that.

kostyanf14:
@yselkowitz We also started PRP for RHEL10. How to handle https://issues.redhat.com/browse/RHELPLAN-171373? Is updating this PR comment with the new bug ID enough?

yselkowitz:
Yes, that's fine.

yselkowitz:
FYI you don't need to keep rebasing for each change that lands before this, just let me know when the package has actually been added.

PR #1348, opened 2025-01-13
Update due to rename of 'mysql' to 'mysql8.4'

FaramosCZ:
Resolves: RHEL-73703


yselkowitz (changes requested):
Unfortunately, Fedora (and hence ELN) has only 8.0, and is currently still unversioned.  Therefore, the configs need to be split, with this become eln-only and unversioned (for now) and a new -c10s config with this applied.

FaramosCZ:
Right, I suspected that.

--

Additional info:
Actually, in Fedora, the package 'mysql' is replaced by 'mysql8.0' too, since F40:
https://src.fedoraproject.org/rpms/mysql8.0

The difference is that the Fedora package provides also the unversioned names, so the content resolver is likely fine with those.

--

To clear the confusion, maybe it would be best that I would create two entirely new workloads, one for 'mysql8.0' for Fedora, listing the versioned names (as there is a plan to dynamically change between Fedora releases which of the different MySQL series in Fedora will be the only one that provides the unversioned names) to be sure they are picked up correctly in any case.

The other workload for 'mysql8.4' for RHEL 10.
The 8.4 is planned for Fedora too, just haven't left Fedora review yet.

At the same time I'd retire the current (unvrsioned) workload entirely. (F39 is EOL already)

--

How does it sound to you?
Any suggestions or reservations?

FaramosCZ:
I implemented the changes I described above,
please re-review @yselkowitz

yselkowitz (approved):

yselkowitz:
To be clear, ELN (like rawhide) has a 8.0 versioned SRPM but the RPMs are still unversioned.  In any case, they provide the versioned names, so this should work for now.  While the RHEL changes need to be finished first due to timing, at some point I would like to get the ELN packages to follow the new pattern by changing [mysqlX.Y.spec](https://src.fedoraproject.org/rpms/mysql8.0/blob/rawhide/f/mysql8.0.spec#_9):

``` diff
@@ -9,1 +9,1 @@
-%{!?mysql_default:%global mysql_default 1}
+%{!?mysql_default:%global mysql_default %{undefined rhel}}
```

And as previously mentioned, perl-DBD-MySQL will need to be updated as well.

PR #1349, opened 2025-01-13
Move copy-jdk-configs to unwanted

gnu-andrew:
Parallel installations of the same JDK are no longer supported and the dependency on copy-jdk-configs has been dropped.

Resolves: RHELMISC-7203

### DIFF
--- a/configs/sst_java-java.yaml
+++ b/configs/sst_java-java.yaml
@@ -17,7 +17,6 @@ data:
     - java-21-openjdk-jmods
     - java-21-openjdk-src
     - java-21-openjdk-static-libs
-    - copy-jdk-configs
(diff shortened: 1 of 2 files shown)


yselkowitz (approved):
The change is fine, but copy-jdk-configs is still a dependency of java-21-openjdk in ELN (as in Fedora), and therefore it will remain in ELN until this dependency is removed.  Will that be part of the implementation of https://fedoraproject.org/wiki/Changes/ThirdPartyLegacyJdks?

gnu-andrew:
This would be for Jiri Vanek to resolve as Fedora maintainer. The dependency has been removed in CentOS 10.

PR #1350, opened 2025-01-14
Update rust components

yselkowitz:
Unversioned wasm32-wasi deprecation:
https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html

Re-enable aarch64-unknown-none-softfloat target:
https://src.fedoraproject.org/rpms/rust/c/f737b8c4dddc2d64481479aec6c85bead138c1af

/cc @cuviper

### DIFF
--- a/configs/sst_pt_llvm_rust_go-rust-toolset.yaml
+++ b/configs/sst_pt_llvm_rust_go-rust-toolset.yaml
@@ -14,11 +14,15 @@ data:
     - rust-lldb
     - rust-std-static
     - rust-std-static-wasm32-unknown-unknown
-    - rust-std-static-wasm32-wasi
     - rust-std-static-wasm32-wasip1
     - rustfmt
     - rust-toolset-srpm-macros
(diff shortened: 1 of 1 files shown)


configs/sst_pt_llvm_rust_go-rust-toolset.yaml
```diff
@@ -14,11 +14,13 @@ data:
...
     - rust-std-static
     - rust-std-static-wasm32-unknown-unknown
-    - rust-std-static-wasm32-wasi
     - rust-std-static-wasm32-wasip1
     - rustfmt
     - rust-toolset-srpm-macros
     - rust-toolset
+  arch_packages:
+    aarch64:
+      - rust-std-static-aarch64-unknown-none-softfloat
```
cuviper, line 23: Do we also need to add `rust-std-static-x86_64-unknown-none`?

cuviper (approved):
LGTM!

PR #1352, opened 2025-01-17
configs/sst_image_builder: add cockpit-image-builder package

croissanne:
The cockpit-image-builder-package will replace cockpit-composer.

Resolves: RHELPLAN-171387

### DIFF
--- a/configs/sst_image_builder-composer.yaml
+++ b/configs/sst_image_builder-composer.yaml
@@ -8,4 +8,4 @@ data:
     - eln
     - c10s
   packages:
-    - cockpit-composer
+    - cockpit-image-builder


yselkowitz (commented):
If it's a replacement, shouldn't `cockpit-composer` be removed?

croissanne:
> If it's a replacement, shouldn't `cockpit-composer` be removed?

The obsoletes is not yet in place, can I already drop it here then or should we wait until it's in place?

ondrejbudai:
@croissanne content resolver is targetting RHEL 11 atm, so it should be fine to remove cockpit-composer already. :)

yselkowitz:
CR covers both c10s and ELN.  Will this apply to c10s (RHEL 10) as well?

croissanne:
> CR covers both c10s and ELN. Will this apply to c10s (RHEL 10) as well?

Yes, but worst case starting 10.1, we're aiming for an obsoletes in 10.0 though.

PR #1354, opened 2025-01-21
Update eln_extras_rhel_packager.yaml

conniebetts:
updating 'maintainer' field in yaml files to align to 'people-tagging' team name

### DIFF
--- a/configs/eln_extras_rhel_packager.yaml
+++ b/configs/eln_extras_rhel_packager.yaml
@@ -3,7 +3,7 @@ version: 1
 data:
   name: RHEL packager
   description: ELN Extras packages for RHEL maintainers
-  maintainer: carlwgeorge
+  maintainer: rhel-sst-cpe
   packages:
     - NetworkManager-openvpn-gnome
     - centpkg


configs/eln_extras_rhel_packager.yaml
```diff
@@ -3,7 +3,7 @@ version: 1
 data:
   name: RHEL packager
   description: ELN Extras packages for RHEL maintainers
-  maintainer: carlwgeorge
+  maintainer: rhel-sst-cpe
```
carlwgeorge, line 6: After the QE merger, the CPE team is now the CLE team.

https://docs.fedoraproject.org/en-US/cle/

Regardless of that, I'm not aware that the CPE team was ever considered a RHEL SST, and the same goes for CLE.  I'm happy to hand off ownership of this to whoever makes the most sense, but it should probably be to a real SST.

conniebetts:
Carl  & Yaakov

I will correct the error I made in making that change.  I apologize for the
confusion it caused.



Connie

yselkowitz (changes requested):
Blocking pending resolution.

carlwgeorge:
What's the end goal here?  This is an ELN Extra workload, so I'm not sure requiring a RHEL SST maintainer is appropriate.  Are we just trying to use groups instead of individuals?

Also pinging @tdawson, who is the main admin for centpkg and is a collaborator for NetworkManager-openvpn, for his input.

yselkowitz:
Thanks @carlwgeorge @tdawson for your input.

PR #1357, opened 2025-01-23
Rename display teams according to team nomenclature (CASE-789)

conniebetts:


yselkowitz:
See #1356

rlemosor:
But it was still a draft :) can we reopen? @conniebetts , we'll use the new nomenclature for the file names, the one Yaakov suggested here:
https://github.com/minimization/content-resolver-input/pull/1356#issuecomment-2613328088
the rename on line #6 (maintainer) remains the same (that's the simplest of the two)

yselkowitz:
This needs to be redone anyway, so let's just start with a clean PR.

PR #1358, opened 2025-01-26
eln_extras_meta: Move azure-cli to separate workload

michel-slm:
This minimizes the blast radius when it fails to build


yselkowitz (approved):
This is fine, but you may want to consider further breaking things out (in additional PR(s)), as your workload often fails.

PR #1359, opened 2025-01-28
Drop sedutil on ppc64le

yselkowitz:
https://src.fedoraproject.org/rpms/sedutil/c/5a431de8d73b061bf53349017149e037b8ec2c96

/cc @davide125 @michel-slm


davide125 (approved):
Thanks!

PR #1366, opened 2025-02-07
Add sdl2-compat

wtay:
See RHELPLAN-171489

### DIFF
--- a/configs/sst_display_hardware_multimedia-graphics-development-libraries.yaml
+++ b/configs/sst_display_hardware_multimedia-graphics-development-libraries.yaml
@@ -8,6 +8,7 @@ data:
     - libdecor
     - SDL
     - SDL2
+    - sdl2-compat
     - SDL3
   labels:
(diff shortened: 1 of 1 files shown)


yselkowitz (changes requested):
hi @wtay thanks for the PR.  Please note that sdl2-compat Provides: SDL2 just like sdl12-compat Provides: SDL.  Therefore, either this can just be left as is (and rely on the Provides), or the -compat packages should replace the original names.

wtay:
> hi @wtay thanks for the PR. Please note that sdl2-compat Provides: SDL2 just like sdl12-compat Provides: SDL. Therefore, either this can just be left as is (and rely on the Provides), or the -compat packages should replace the original names.

Ok, then I think it's better to leave it as is for now.

Eventually we want to remove the SDL2 package and only have sdl2-compat. When that happens, we should replace the SDL2 line with sdl2-compat.

PR #1367, opened 2025-02-07
Fix unwantedness of pulseaudio

yselkowitz:
As noted in the comment, the pulseaudio SRPM is not completely unwanted; the various libs subpackages are still needed, while the server and modules are not, and in fact are no longer built for RHEL.

/cc @nielsdg


nielsdg:
Probably @wtay is a better person to review this

yselkowitz:
ping @wtay

wtay:
This is correct. We don't want the server part and its modules anymore but we still want the -libs for pulseaudio.

PR #1368, opened 2025-02-07
Rename display teams according to new team nomenclature (CASE-789 & 894)

conniebetts:
Rename display teams according to new team nomenclature (CASE-789 & 894)


yselkowitz (changes requested):
First off, this needs to be rebased against the current `main` branch in order to resolve the merge conflict.

Some of these are mistakenly using underscore as a filename separator instead of the agreed-upon double hyphen.

Finally, please include the SST name in the commit messages, so that we don't end up with dozens of commits with the same title.

conniebetts:
@yselkowitz thank you, I will make that correction once I've reviewed the documentation and steps on making the correction.

yselkowitz:
This is being done in #1395.

PR #1369, opened 2025-02-08
Update for tracker->tinysparql and tracker-miners->localsearch

AdamWill:

PR #1370, opened 2025-02-09
Revert "Update for tracker->tinysparql and tracker-miners->localsearch"

yselkowitz:
This was correct for ELN but not for C10S.  Now that tinysparql/localsearch has been built for ELN, they will still be used as they provide the old tracker names.

Reverts minimization/content-resolver-input#1369

/cc @AdamWill @tpopela


tpopela:
Thank you @yselkowitz !

PR #1373, opened 2025-02-13
Drop plasma-applet-translator

yselkowitz:
This is a Plasma 5 QML applet which cannot be compatible with Plasma 6 and has not had an upstream release in years.

This should probably be retired from Fedora and EPEL 10 (!) as well.

/cc @tdawson


yselkowitz:
ping @tdawson

tdawson:
You are correct.  We are not planning on shipping qt5 in EPEL11, so yes, if this looks like it's not going to be ported to qt6, then let's drop this.

PR #1374, opened 2025-02-17
List icu, add compat upgrade helper

yselkowitz:
This is intended to avoid the usual breakage in ELN when icu gets a major version bump.

/cc @juhp


juhp:
Thanks!

cc @erack and @mike-fabian

PR #1375, opened 2025-02-18
Reflect the sst split of sst-virt-tools from sst-virtualization in Co…

yantian0:
…ntent Resolver


yselkowitz (commented):
Is `python3-libvirt` left behind on purpose?

yantian0:
> Is `python3-libvirt` left behind on purpose?

Fixed, thanks for your point, Yaakov.

yselkowitz (approved):

PR #1376, opened 2025-02-18
thunderbird is not build on s390x

tdawson:
thunderbird is not build on s390x so make everything arch_packages.

### DIFF
--- a/configs/sst_display_productivity-thunderbird.yaml
+++ b/configs/sst_display_productivity-thunderbird.yaml
@@ -4,8 +4,14 @@ data:
   name: Mozilla Thunderbird
   description: Default e-mail client for Workstation
   maintainer: sst_display_productivity
-  packages:
-    - thunderbird
+  packages: []
+  arch_packages:
+    aarch64:
+      - thunderbird
+    ppc64le:
+      - thunderbird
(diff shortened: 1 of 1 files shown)


yselkowitz:
@tdawson but it is for RHEL s390x, e.g. https://kojihub.stream.centos.org/koji/buildinfo?buildID=74808 so I had been hoping the fix would be coming back to Fedora (and hence ELN Extras) at some point.

@tpopela any insight on this?

tpopela:
@yselkowitz I don't, but to me it looks life a forgotten change (done 4 years ago - https://src.fedoraproject.org/rpms/thunderbird/c/c3cb08572fafaafbf229645ff3841151567ea543). I will sync with FF/TB team. Thank you for the heads-up!

tpopela:
@yselkowitz I don't, but to me it looks life a forgotten change (done 4 years ago - https://src.fedoraproject.org/rpms/thunderbird/c/c3cb08572fafaafbf229645ff3841151567ea543). I will sync with FF/TB team. Thank you for the heads-up!

yselkowitz:
TB is now being built for s390x in both Fedora and ELN.

PR #1378, opened 2025-02-21
meta: split off hardware related packages

michel-slm:
Move these from the main workload to a `meta_hw` one

Also start tracking `aer-inject` and `mce-inject`

### DIFF
--- a/configs/eln_extras_meta.yaml
+++ b/configs/eln_extras_meta.yaml
@@ -37,9 +37,7 @@ data:
     - libidn-devel
     - libkdumpfile-devel
     - libxcrypt-compat
-    - linux-sysinfo-snapshot
     - mkosi
-    - mlnx-tools
     - mosh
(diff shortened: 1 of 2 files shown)


configs/eln_extras_meta.yaml
```diff
@@ -39,7 +39,6 @@ data:
     - libxcrypt-compat
     - linux-sysinfo-snapshot
```
yselkowitz, line 40: Did you mean to remove this from here while adding it to the new config?

yselkowitz (changes requested):

configs/eln_extras_meta.yaml
michel-slm: uh... thanks, yes I did

michel-slm: updated, thanks

yselkowitz (approved):

PR #1382, opened 2025-02-27
Update pcs dependencies to match the current state

CtrlZmaster:
- we unbundled Tornado in RHEL 9.6&10.0GA+
- findutils and tar were used before, just not declared
- systemd-rpm-macros is now used instead of systemd per Fedora Packaging Guidelines example for systemd scriptlet macros

@yselkowitz I have a question about what actually goes in this file... Is it just the BuildRequires declared in the spec file? Or the output of `rpm -q --requires <rpm>` (for Requires) and `rpm -q --requires <srpm>` (for BuildRequires)? There is a difference since the latter also contains dependencies automatically generated by RPM.

And what about `BuildRequires: pkgconfig(booth)`? I changed to this style of dependency to indicate that we only need the pkgconfig of the package and to make sure if the pc file is ever split from the main package, we won't have to make changes. Does it belong here as a resolved rpm name or should I stick to exact copy from the spec file?

I have moved away from declaring [Build]Require `python3` version explicitly in the spec file since rpm generates the correct dependency automatically and Python packages should just BuildRequire `python-devel`. However, I left `python3` in the workload file because it feels wrong removing it.

### DIFF
--- a/configs/sst_high_availability-userspace.yaml
+++ b/configs/sst_high_availability-userspace.yaml
@@ -168,8 +168,8 @@ data:
         - automake
         - booth
         - coreutils
+        - corosync-devel
         - corosync-qdevice-devel
-        - corosynclib-devel
         - diffstat
         - findutils
         - gcc
@@ -177,13 +177,13 @@ data:
         - git-core
         - libappstream-glib
         - libcurl-devel
+        - libffi-devel
         - make
         - nodejs-npm
         - nss-tools
         - openssl-devel
         - pacemaker-libs-devel
         - pam
-        - python3
         - pkgconfig
         - python3-cryptography
         - python3-dateutil
@@ -214,12 +214,15 @@ data:
           dependencies:
             - corosync
             - libcurl
+            - libffi
             - libknet1-plugins-all
             - logrotate
             - nss-tools
+            - openssl-libs
             - pacemaker-cli
             - pam
             - pcmk-cluster-manager
+            - pkgconfig
             - psmisc
             - python3
             - python3-cryptography
--- a/configs/sst_high_availability-userspace.yaml
+++ b/configs/sst_high_availability-userspace.yaml
@@ -171,6 +171,7 @@ data:
         - corosync-qdevice-devel
         - corosynclib-devel
         - diffstat
+        - findutils
         - gcc
         - gcc-c++
         - git-core
@@ -183,6 +184,7 @@ data:
         - pacemaker-libs-devel
         - pam
         - python3
+        - pkgconfig
         - python3-cryptography
         - python3-dateutil
         - python3-devel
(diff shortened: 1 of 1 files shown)


yselkowitz (approved):
> * we unbundled Tornado in RHEL 9.6&10.0GA+
>     * findutils and tar were used before, just not declared
>     * systemd-rpm-macros is now used instead of systemd per Fedora Packaging Guidelines example for systemd scriptlet macros

These changes look fine.

> @yselkowitz I have a question about what actually goes in this file... Is it just the BuildRequires declared in the spec file? Or the output of `rpm -q --requires <rpm>` (for Requires) and `rpm -q --requires <srpm>` (for BuildRequires)? There is a difference since the latter also contains dependencies automatically generated by RPM.

The latter.  A placeholder undergoes no further processing, so the information provided here needs to be complete.

> And what about `BuildRequires: pkgconfig(booth)`? I changed to this style of dependency to indicate that we only need the pkgconfig of the package and to make sure if the pc file is ever split from the main package, we won't have to make changes. Does it belong here as a resolved rpm name or should I stick to exact copy from the spec file?

For spec files you should absolutely follow the packaging guidelines, but literal package names should be used in CR.  (See https://github.com/fedora-eln/eln/issues/212 for more details about virtual provides.)

> I have moved away from declaring [Build]Require `python3` version explicitly in the spec file since rpm generates the correct dependency automatically and Python packages should just BuildRequire `python-devel`. However, I left `python3` in the workload file because it feels wrong removing it.

`python3` could be removed from `build_dependencies` but left in `dependencies` in this case.

configs/sst_high_availability-userspace.yaml
```diff
@@ -210,19 +214,22 @@ data:
...
             - libcurl
+            - libffi
             - libknet1-plugins-all
             - logrotate
             - nss-tools
+            - openssl-libs
             - pacemaker-cli
             - pam
             - pcmk-cluster-manager
+            - pkgconfig
```
yselkowitz, line 225: `pkgconfig` is a *runtime* dependency?

yselkowitz (changes requested):

configs/sst_high_availability-userspace.yaml
CtrlZmaster: I explained it in the commit message. It was in the output of `rpm -q --requires pcs`. I was surprised too.

yselkowitz: Oh I see, the pcs package includes `pcs.pc` pkgconfig metadata (applies to Fedora as well).  That's unusual (but not completely unheard of) for a runtime package.  While you may want to consider separating that into a -devel subpackage, I'm not sure what else you would put into that, and first you'd need to convert any `BuildRequires: pcs` accordingly.  Something to consider, but not urgent.

yselkowitz (approved):

PR #1384, opened 2025-02-28
Drop geolite2 databases from ELN

yselkowitz:
These free versions are not planned to be provided in RHEL 11; users will need to get the databases themselves (often commercially).  libmaxminddb remains as a bind dependency for that purpose.


mruprich:
LGTM! Thanks!

PR #1385, opened 2025-02-28
Revert "thunderbird is not build on s390x"

yselkowitz:
thunderbird is being built again on s390x in both Fedora and ELN Extras.

Reverts minimization/content-resolver-input#1376

/cc @tdawson @tpopela


tpopela:
Yes, Jan fixed the build. Thank you!

PR #1388, opened 2025-03-03
Rename teams according to new nomenclature (CASE-789)

rlemosor:


configs/rhel-sst-cs-base-utils--management-tog-pegasus.yaml
```diff
@@ -3,7 +3,7 @@ version: 1
 data:
```
yselkowitz, line 3: The `sys-` in sys-management was dropped from the filename here.

yselkowitz (changes requested):

configs/rhel-sst-cs-base-utils--management-tog-pegasus.yaml
rlemosor: Good catch! I'll find a way to do this with a local script instead of renaming the files online with github.dev.

yselkowitz: Would it make more sense if I script this and you review the results?

rlemosor: Think I managed to get a script working, please check the results -- but thanks for the offer!

rlemosor:
As Yaakov noted, my scripted operation was removing the old files and adding new ones in its stead, breaking history; which is unwanted. So we're dropping this one in favor of https://github.com/minimization/content-resolver-input/pull/1395/files

PR #1390, opened 2025-03-04
kde: enable cantor, kaidan; enable qtwebengine on ppc64le

yselkowitz:


tdawson:
Looks good.  Thanks for this.

PR #1393, opened 2025-03-11
Rename network-fastdatapath to network-fastdatapath-ovsdpdk

smwfaye:
Rename network-fastdatapath to network-fastdatapath-ovsdpdk as part of the splitting of the network fastdatapath SST.


yselkowitz (approved):
Normally, we rename the files themselves when changing or renaming SSTs, but we're about to do a mass renaming anyway which will catch this.

PR #1394, opened 2025-03-11
Add missing sub-packages for gnome-kiosk

Jiri-Koten:
gnome-kiosk-script-session and gnome-kiosk-search-appliance are shipped in AppStream and therefore should be part of the workload.

Also required for https://github.com/fedora-eln/eln/issues/228.

### DIFF
--- a/configs/sst_display_window_management-gnome-kiosk.yaml
+++ b/configs/sst_display_window_management-gnome-kiosk.yaml
@@ -6,6 +6,8 @@ data:
   maintainer: sst_display_window_management
   packages:
     - gnome-kiosk
+    - gnome-kiosk-script-session
+    - gnome-kiosk-search-appliance
   labels:
     - eln
     - c10s


yselkowitz (changes requested):
Doing this as things currently stand would cause firefox to be pulled into ELN, which is supposed to be flatpak-only in RHEL 11.  Also, the Fedora build thereof is still using python3.11 to build, which is also unwanted in ELN.

Possible remedies include:
- Patching gnome-kiosk to use a firefox flatpak instead of an RPM (as in RHEL 10), but then the question will be what will provide such a flatpak in/for ELN;
- Maintaining an `eln` branch of firefox based on `c10s`, and building that as an RPM;
- (Conditionally) bundle any unwanted dependencies in the Fedora firefox RPM builds, and start including the ELN builds therefrom

All of these will require work from the Display team(s) though.

PR #1398, opened 2025-03-24
Adapt the Fedora 41 Change, ibus-chewing default IM for zh_TW locale

pnemade:
https://fedoraproject.org/wiki/Changes/IBusChewingForZhTW


pnemade:
Also reported https://pagure.io/fedora-comps/pull-request/1105

yselkowitz (changes requested):
The original config should remain ELN, and any c10s specific settings in a `-c10s` suffixed config.

PR #1400, opened 2025-03-26
Add `postgis` extension for postgresql and it's dependencies

ljavorsk:
Resolves: RHELPLAN-171698 RHELPLAN-171818 RHELPLAN-171821 RHELPLAN-171697


yselkowitz (changes requested):
Thank you for your PR.  Some of these components are going to require changes first in (RHEL and) ELN in order to avoid unwanted dependencies, so this can't be merged quite yet, but you may proceed with the PRP in the meantime.

- proj has some `deprecated()` subpackages which presumably should be disabled for RHEL and ELN.  It's mingw subpackages would already be disabled.
- geos should have its mingw subpackages disabled when built for RHEL; if wanted in EPEL, then the separate `mingw-geos` package should be revived instead.
- gdal has a **LOT** of additional dependencies in Fedora.  Some of those may ultimately be optional, but will need to be conditionalized.
- postgis in Fedora also has a dependency on SFCGAL, will that be disabled in RHEL?
- Will the postgis SRPM be renamed to e.g. postgresql16-postgis like other postgresql-related components to allow future versions per the latest naming scheme?

fila43:
Hi @yselkowitz, 
I hope we have already addressed all of your comments in Fedora's PRs. Any kind of review is welcome.

- [gdal](https://src.fedoraproject.org/rpms/gdal/pull-request/23)
- [geos](https://src.fedoraproject.org/rpms/geos/pull-request/4)
- [proj](https://src.fedoraproject.org/rpms/proj/pull-request/11)
- [postgis](https://src.fedoraproject.org/rpms/postgis/pull-request/8)

Yes, we plan to use the versioned name for Postgis, as we already do for the other postgresql extensions. Here, I am not sure how to approach content resolver because in Fedora, there is going to be a postgis component, but in RHEL, there will be postgresql16-postgis. Is such a difference acceptable?

yselkowitz:
Thanks for the PRs, I'm following those now.

As for postgis vs. postgresql16-postgis, since the difference is only in the SRPM name while CR tracks RPMs, it's okay for now, but it will make things difficult when we start preparing c11s.  Therefore, while you should focus on getting these ready for RHEL 10 and ELN through the PRs and PRP etc., once everything is set, then we'll need to look at renaming the package in Fedora to match the new naming scheme.

fila43:
@yselkowitz  So, can we merge this PR for now?

yselkowitz:
No, because merging this would cause those packages to be added to ELN, and in their current state, that would add a bunch of unwanted dependencies to ELN.  You may proceed with the RHEL PRP in the meantime.

fila43:
Ok, I will ping you once the MRs in Fedora are done.

fila43:
@yselkowitz We have discussed a renaming scheme for PostGIS in my team.  Nowadays, all renamed components must be recreated in Fedora for every single Postgresql version. It consumes our capacity and the profit for Fedora is questionable. What is your opinion on keeping it as it is without the version? It would continue as a rolling stream in Fedora, and once c11s comes, we will rename it there.

yselkowitz (approved):
The packaging changes have all landed in rawhide and ELN.

The SRPM name doesn't need to be fixed today, but it should be fixed eventually not only for consistency with the rest of Fedora, and because it will affect the import of RHEL 11.  It looks like postgresql17 packages are on their way, that would provide a good opportunity to get this on the same naming scheme as the rest of the packages.

PR #1401, opened 2025-03-27
sst-cockpit: Drop cockpit-ostree

martinpitt:
RHEL 4 Edge was only a RHEL 9 thing, and does not exist in RHEL 10 any more. This extension is undermaintained, seems to have no significant user base, and does not make much sense with bootc images. So drop it for RHEL 11.

### DIFF
--- a/configs/rhel-sst-cockpit--web-console-c10s.yaml
+++ b/configs/rhel-sst-cockpit--web-console-c10s.yaml
@@ -5,7 +5,7 @@ data:
   description: Cockpit basic packages
   maintainer: rhel-sst-cockpit
   labels:
-    - eln
+    - c10s
   packages:
     # meta-package for the basic stuff (web server, bridge, core pages)
     - cockpit
@@ -13,6 +13,8 @@ data:
     - cockpit-packagekit
     - cockpit-storaged
     - cockpit-doc
+    # RHEL 4 Edge image management
+    - cockpit-ostree
     # extensions maintained by Cockpit team
     - cockpit-files
   arch_packages:


yselkowitz (changes requested):
@martinpitt thanks for your PR.  Will cockpit-ostree still be shipped in RHEL 10?  If so, we'll need to continue to track it, but for c10s only.  To do that, in addition to this change, you'll need to add a config (e.g. `rhel-sst-cockpit--web-console-c10s.yaml` with only `cockpit-ostree` listed and only the `c10s` label.

martinpitt:
> Will cockpit-ostree still be shipped in RHEL 10?

I suppose we have to -- it's in 10.0, so we sort of (accidentally) committed to it.

I did the suggested change. I also added a "c11s" label to the "main" file. If that is too early, I'm of course happy to revert that. Thanks!

configs/rhel-sst-cockpit--web-console.yaml
```diff
@@ -6,16 +6,14 @@ data:
   maintainer: rhel-sst-cockpit
   labels:
     - eln
-    - c10s
+    - c11s
```
yselkowitz, line 9: `c11s` is not yet a valid label.  The way you split this, just drop cNs labels and leave it `eln` only.  When C11S is branched, then all workloads with an `eln` label will have a `c11s` label added.
```suggestion
```

yselkowitz (changes requested):

configs/rhel-sst-cockpit--web-console.yaml
martinpitt: Ack! updated.

yselkowitz (approved):

hroncok:
Followup: https://pagure.io/pungi-fedora/pull-request/1469

PR #1407, opened 2025-04-01
insights-client should not runtime-require python3-setuptools

hroncok:
See https://github.com/RedHatInsights/insights-client/pull/397


yselkowitz (approved):

hroncok:
Note that this is still a draft.

hroncok:
Ready.

PR #1408, opened 2025-04-01
meta-sig: sequoia-keyring-linter was retired

yselkowitz:
The replacement is `sq cert lint`:

https://src.fedoraproject.org/rpms/rust-sequoia-keyring-linter/c/94784a2dbe293ceefba3d1cd24150f96d7a734cf

/cc @davide125 @michel-slm


michel-slm:
ack, thank you

PR #1409, opened 2025-04-01
Meta: drop atop, add colorized-logs

michel-slm:
We use below instead of atop

Also track colorized-logs, useful for dealing with output of tools that do not intelligently strip ANSI formatting when piped to other tools


yselkowitz (approved):
"We use below instead of atop" it admittedly took me a minute to get that...  just gotta love creating software names.

PR #1416, opened 2025-04-25
Add new hunspell-ka package in ELN

pnemade:
This PR will add new hunspell-ka package in RHEL 11.


yselkowitz:
You don't have any plans to add this to RHEL 10.y?

pnemade:
I do have plan, but I have not yet started PRP. Also not sure if it will take 1 week for me in the process to report to add it to RHEL 10.y. Hence this PR.

So suggest next steps.

yselkowitz:
In that case, just add `hunspell-ka` to the main config, there is no need to split the configs for a few weeks delay.

pnemade:
Please review again and merge this PR

yselkowitz (approved):

PR #1421, opened 2025-05-19
Add new libguestfs-fssupport package (RHEL only)

rwmjones:
Related: https://issues.redhat.com/browse/RHEL-92064

### DIFF
--- a/configs/rhel-sst-virt-tools--subsystem.yaml
+++ b/configs/rhel-sst-virt-tools--subsystem.yaml
@@ -44,6 +44,45 @@ data:
       - virt-top
   package_placeholders:
     # RHEL-specific package, not in Fedora
+    - srpm_name: libguestfs-fssupport
+      build_dependencies:
+        - autoconf
+        - automake
+        - e2fsprogs-devel
+        - elfutils-libelf-devel
+        - gcc
+        - gnupg2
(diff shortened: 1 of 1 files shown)


configs/rhel-sst-virt-tools--subsystem.yaml
```diff
@@ -44,6 +44,45 @@ data:
...
+        - kmod
+        - libacl-devel
+        - lzo-devel
+        - make
+        - pkgconfig(blkid)
+        - pkgconfig(libgcrypt)
+        - pkgconfig(libudev)
+        - pkgconfig(libzstd)
+        - pkgconfig(uuid)
+        - pkgconfig(zlib)
```
yselkowitz, line 67: Package names need to be used for placeholders.

yselkowitz (changes requested):

configs/rhel-sst-virt-tools--subsystem.yaml
rwmjones: Why do we even need this information here?

(And if you're going to tell me it needs to be up to date, then dozens, probably hundreds of existing packages have wrong build dependencies here.)

yselkowitz (changes requested):

yselkowitz (approved):

PR #1423, opened 2025-05-29
hyperscale: drop fsverity-utils

yselkowitz:
This has yet to be ported away from OpenSSL ENGINE and has not been built in EPEL 10 or CBS as a result.

/cc @michel-slm per discussion in ELN SIG today.


michel-slm:
Ack, this should be fine to merge. I'll write a matching change for our internal package tracker.

PR #1425, opened 2025-06-05
Ship new kernel-modules-extra-matched meta package

SirPhuttel:
Related: https://issues.redhat.com/browse/RHEL-87452

### DIFF
--- a/configs/rhel-sst-kernel-maintainers--kernel-base.yaml
+++ b/configs/rhel-sst-kernel-maintainers--kernel-base.yaml
@@ -39,6 +39,7 @@ data:
     - kernel-modules
     - kernel-modules-core
     - kernel-modules-extra
+    - kernel-modules-extra-matched
     - kernel-tools
     - perf
     - python3-perf


yselkowitz:
@SirPhuttel thank you for the PR.  Looking at koji, this subpackage is built both archful and noarch, which doesn't seem right.  Could you please verify?

SirPhuttel:
@yselkowitz good catch! I asked about it in RHEL-87452. IMO it should suffice to be noarch only, as it a) has no arch-specific content and b) does not directly pull in kernel-modules-extra package but instead kernel package has a conditional dependency on it:
`Requires: ((kernel-modules-extra-uname-r = %{KVERREL}) if kernel-modules-extra-matched)`
But is that detail relevant for this MR? The list I'm adding kernel-modules-extra-matched to contains both archful packages (e.g. kernel-modules-extra) and noarch ones (e.g. kernel-doc). So whether extra-matched should be noarch or not, it's still added to the same spot, right?

PR #1426, opened 2025-06-06
Change wsl-setup to eln

tdawson:
wsl-setup is now being added to RHEL 9/10.
Change config to move it from eln-extras to eln


hroncok:
@tdawson the package has no owner in RHEL 11. Could you please follow up by creating the owner file in bugzilla-data/components with the assigned maintainer, QA and SST? Thanks.

Ping me on slack if you want internal links.

PR #1427, opened 2025-06-10
Introduce Java RPM Packaging Tools workload

mizdebsk:
Add a new workload that provides tools for packaging Java software as RPMs.  On ELN, it supports both the new declarative and traditional imperative build systems; on C10S, only the traditional imperative approach is supported.

This new workload supersedes the legacy java-bootstrap workload.


yselkowitz (approved):
Looks like this will add a few additional dependencies, hopefully reasonable, but I'll follow up if it's more than expected.

PR #1438, opened 2025-06-25
Add insights-core-selinux to insights-client dependencies

xiangce:
- insights-core-selinux is a new package that will be added to RHEL 9.7/10.1. It's required when installing the insights-client.
- Jira: RHELPLAN-172287

### DIFF
--- a/configs/rhel-sst-csi-client-tools--all.yaml
+++ b/configs/rhel-sst-csi-client-tools--all.yaml
@@ -53,6 +53,13 @@ data:
           dependencies:
             - subscription-manager
             - systemd
+    - srpm_name: insights-core-selinux
+      rpms:
+        - rpm_name: insights-core-selinux
+          description: insights-core-selinux
+          dependencies:
+            - selinux-policy
+            - selinux-policy-targeted
   labels:
     - eln
     - c10s


yselkowitz:
If this is going to be another RHEL-specific package, it needs to be added as a new placeholder instead.

xiangce:
@yselkowitz - I added it to a new placeholder as well, please take a look again. Thanks.

configs/rhel-sst-csi-client-tools--all.yaml
```diff
@@ -22,6 +22,7 @@ data:
           dependencies:
             - coreutils
             - gpg
+            - insights-core-selinux
```
yselkowitz, line 25: Drop the dependency here, CR doesn't handle placeholders as dependencies.

xiangce: Hi @yselkowitz - What we want to do is to ensure the new `insights-core-selinux` one of the dependencies of the `insights-client`. (i.e., when installing `insights-client` package, the `insights-core-selinux` should be installed at the same time). Will removing this line from the `dependencies` break this dependency relationship? 

Sorry for the first grade question.

yselkowitz: You will still need the dependency in the spec file, that is what actually matters.  CR can't currently represent that though, so it needs to be omitted here.

xiangce: Got it. Just removed it. Thanks for the clarification.

yselkowitz (approved):

PR #1439, opened 2025-06-26
Try to add a base set of Fedora rawhide packages used for i686

smooge:
Using the steam package spec file and a general list of what might be useful for i686 compilation, I have brutally copied and pasted from other config files what looks like the minimum needed for making useful content resolver files.

### DIFF
--- a/configs/fedora-sig-i686-compilers.yaml
+++ b/configs/fedora-sig-i686-compilers.yaml
@@ -1,0 +1,63 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: C Development Environment for i686
+  description: A C development environment for i686
+  maintainer: ssmoogen
+  packages:
+    # The minimum development environment is this:
+    - gcc
+    - binutils
+    - glibc
+    - glibc-common
+    - glibc-devel
+    # Full development includes additional packages:
+    - glibc-utils
+    - glibc-static
+    - glibc-nss-devel
+    # Support all locales.
+    - glibc-all-langpacks
+    # parser and scanners
+    - bison
+    - bison-devel
+    - bison-runtime
+    - byacc
+    - flex
+    - flex-devel
+    # LLVM environment
+    - llvm
+    - llvm-devel
+    - llvm-googletest
+    - llvm-libs
+    - llvm-static
+    - llvm-test
+    - llvm-toolset
+    - clang
+    - clang-devel
+    - clang-libs
+    - clang-tools-extra
+    - clang-resource-filesystem
+    - git-clang-format
+    - python3-clang
+    - compiler-rt
+    - libomp
+    - libomp-devel
+    - lld
+    - lld-devel
+    - lld-libs
+    - lldb
+    - lldb-devel
+    - python3-lldb
+    - python3-lit
+    # make
+    - make
+    - make-devel
+    # autotools
+    - autoconf
+    - autoconf-archive
+    - automake
+    - libtool
+    - m4
+
+  labels:
+    - fedora-rawhide-i686
--- a/configs/fedora-sig-i686-steam.yaml
+++ b/configs/fedora-sig-i686-steam.yaml
@@ -1,0 +1,43 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: C Development Environment for i686
+  description: A C development environment for i686
+  maintainer: ssmoogen
+  packages:
+    # The minimum development environment is this:
+    - desktop-file-utils
+    - libappstream-glib
+    - make
+    - systemd
+    - tar
+    - zenity
+    # Requirements from https://github.com/rpmfusion/steam/blob/master/steam.spec
+    - NetworkManager-libnm
+    - SDL2
+    - alsa-lib
+    - alsa-plugins-pulseaudio
+    - firewalld-filesystem
+    - fontconfig
+    - gtk2
+    - libICE
+    - libXScrnSaver
+    - libXext
+    - libXinerama
+    - libXtst
+    - libatomic
+    - libcurl
+    - libnsl
+    - libpng
+    - libva
+    - libvdpau
+    - mesa-dri-drivers
+    - mesa-libEGL
+    - mesa-libGL
+    - mesa-vulkan-drivers
+    - nss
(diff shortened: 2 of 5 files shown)


tdawson:
The running version of Content Resolver does not do i686.

bookwar:
It is true that Content Resolver for now does not support dependency resolution for i686 packages (they are mixed into x86_64 repo and it creates issues for the tool). But we can use it to specify the i686-relevant *workloads*.

Even though the dependencies will be resolved against x86_64 packages, the dependency tree produced can be used as a first order approximation of the dependency tree of i686-only packages.

I would suggest naming the base- and the repository- file in a more generic way as they don't really have anything to do with i686 arch.

But then the *workloads* we define will be the i686 workloads.

Having them collected will already help us, even though the full i686 support in Content Resolver is not yet available.

configs/fedora-sig-i686-compilers.yaml
```diff
@@ -0,0 +1,63 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: C Development Environment for i686
+  description: A C development environment for i686
```
yselkowitz, line 5: This file is unnecessary; CR will include any necessary compilers as part of the buildroot dependencies.

configs/fedora-sig-i686-steam.yaml
```diff
@@ -0,0 +1,46 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Packages Required By Steam Spec File
+  description: Packages Required By Steam
+  maintainer: ssmoogen
+  packages:
+    # The minimum development environment is this:
```
yselkowitz, line 8: BuildRequires from https://github.com/rpmfusion/steam/blob/master/steam.spec

configs/fedora-sig-i686-steam.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - desktop-file-utils
+    - libappstream-glib
+    - make
+    - systemd
+    - tar
+    - zenity
```
yselkowitz, line 14: These are Requires, not BuildRequires, but are programs and therefore the x86_64 builds would suffice.

configs/fedora-sig-i686-steam.yaml
```diff
@@ -0,0 +1,46 @@
...
+    # Requirements from https://github.com/rpmfusion/steam/blob/master/steam.spec
+    - NetworkManager-libnm
```
yselkowitz, line 16: All we would need to consider here are those requires which are `%{?_isa}`-suffixed, and therefore they could be listed here as `NetworkManager-libnm.i686` and so on.

configs/fedora-sig-i686-steam.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - SDL2
```
yselkowitz, line 17: SDL2 is provided by sdl2-compat nowadays, so this should be `sdl2-compat.i686`.

configs/fedora-sig-i686-steam.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - alsa-lib
+    - alsa-plugins-pulseaudio
+    - firewalld-filesystem
```
yselkowitz, line 20: This is a noarch subpackage, so it should not need to be included here.

configs/fedora-sig-i686-steam.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - mesa-dri-drivers
+    - mesa-libEGL
+    - mesa-libGL
+    - mesa-vulkan-drivers
+    - nss
+    - pipewire-alsa
+    - pulseaudio-libs
+    - systemd-libs
+    - vulkan-loader
+    - xdg-desktop-portal
```
yselkowitz, line 43: xdg-desktop-portal is a program, so the native x86_64 version should suffice, and this can be omitted.

gamemode.i686 is missing though.

configs/view-fedora-rawhide-i686.yaml
```diff
@@ -0,0 +1,12 @@
+---
+document: feedback-pipeline-compose-view
+version: 1
+data:
+  name: Fedora Rawhide Package Set for i686
+  description: Fedora Rawhide Package Set for i686
+  maintainer: ssmoogen
+  labels:
+  - fedora-i686
```
yselkowitz, line 9: Labels need to be used consistently, as that is what ties configs together.

yselkowitz (changes requested):

configs/fedora-sig-i686-compilers.yaml
bookwar: I think we should include it, because i686 compilers support is needed not only as a dependency for Steam, but as a standalone workload.

And while it should overlap with Steam needs it does not necessarily coincide.

configs/fedora-sig-i686-steam.yaml
bookwar: +1 

But also we should assume that there will be a lot more tweaking and adjustments once we get the initial workload definition publicly visible. Thus I would not be too picky for the items on the current list.

I think that the first step is to get _some_ definition of the workload in, publish it, and then invite people to scrutinize it further.

configs/fedora-sig-i686-steam.yaml
smooge: Added. I thought it was a rpmfusion package when I looked.

configs/fedora-sig-i686-steam.yaml
smooge: removed

configs/fedora-sig-i686-compilers.yaml
smooge: I was going from the tools team people who said they were wanting stuff to be able to debug i686 issues. i686 is still a target for embedded processing and Fedora was what they were currently using to do that. How can I better capture that need?

configs/fedora-sig-i686-steam.yaml
smooge: I think I have fixed this finally by cutting it down to just the packages with %{_isa} in them.

PR #1440, opened 2025-06-27
RHELMISC-13989: mingw-qemu-ga-win deps update

kostyanf14:


configs/rhel-sst-virtualization-windows--subsystem.yaml
```diff
@@ -13,14 +13,15 @@ data:
     - srpm_name: mingw-qemu-ga-win
       build_dependencies:
         - libtool
+        - zlib-devel
```
yselkowitz, line 16: Why revert this?  This is now provided by `zlib-ng-compat-devel`.

yselkowitz (changes requested):
Question below.  msitools will also need to be modified to not BuildRequires this (for testing), but I can handle that.

PR #1453, opened 2025-07-15
Add secvarctl to el9 and el10

marta-lewandowska:
secvarctl is a suite of tools to manage Secure Boot variables on POWER, and is needed for dynamic key management, which is finally being added to RHEL.


yselkowitz:
@marta-lewandowska thanks for the PR.  This package is built for all architectures, but should it be shipped therefor?  What purpose would this serve on a machine other than POWER?

marta-lewandowska:
ok, done :)

yselkowitz (approved):
When completing other steps in the PRP (e.g. product listings/pungi configs), be sure to specify that this is intended for ppc64le only.

PR #1456, opened 2025-07-16
Update rhel-sst-security-crypto--openssl.yaml

simo5:


yselkowitz (changes requested):
AFAICS this is intended to be a separate source package, so there needs to be a new `srpm_name` entry with this under it.

simo5:
@yselkowitz sorry about that, updated patch

yselkowitz (approved):

PR #1466, opened 2025-07-28
Mark python3-wheel unwanted in ELN

hroncok:
See https://github.com/fedora-eln/eln/issues/284

---

Note that the package is listed in configs/rhel-sst-high-availability--userspace.yaml as a build dependency of pcs. I don't know if the package can be wanted and unwanted at the same time, but we don't want it.


yselkowitz (approved):
In this respect, a placeholder BR is no different than any other, so this is okay.

PR #1467, opened 2025-07-28
Add python3-build to desired ELN Python CRB packages

hroncok:
See https://github.com/fedora-eln/eln/issues/278#issuecomment-3094715629


yselkowitz:
@hroncok just to confirm, even though nothing actually needs it yet?  The couple of packages that pulled it in last week didn't actually need `build` in the package building process, it was used for developer/maintainer targets only, and the dependencies have since been removed.

hroncok:
Indeed even then. See the comment above the list. Maintaining this package in CRB for 20 years is less work than adding it to RHEL post GA.

hroncok:
https://pagure.io/fedora-comps/pull-request/1140

PR #1468, opened 2025-07-28
Add dotnet10.0 (.NET 10.0) placeholder

omajid:


yselkowitz:
Why can't this be added to Fedora at least at the same time as RHEL, if not sooner?

omajid:
> Why can't this be added to Fedora at least at the same time as RHEL, if not sooner?

I am working on that. It will take a few weeks. But the RHEL schedule really wants me to open this PR before I can proceed with the RHEL package addition at all.

yselkowitz:
This PR is open, so you may proceed with the PRP, but it should be a real package, so this will wait until it's available.

Also, iiuc 10.0 should be an LTS?  If so, then 8.0 and 9.0 should be dropped from ELN, while remaining in C10S for the remainder of their lifetimes.

omajid:
https://bugzilla.redhat.com/show_bug.cgi?id=2397031

yselkowitz:
#1501 may be a better way of handling this.

PR #1472, opened 2025-08-01
chore: add insights-core-selinux to insights-core dependencies

xiangce:
- the insights-core-selinux should be installed as well
  when installing the insights-core RPM package


yselkowitz:
Content Resolver does not handle placeholder dependencies of other placeholders properly, nor does it serve much purpose to do so.  This of course does not contradict adding the dependencies in the actual spec files, of course.

PR #1474, opened 2025-08-04
Split rhel-sst-pt-python-ruby-nodejs into rhel-pt-python and rhel-pt-ruby-nodejs

hroncok:
Related to https://github.com/fedora-eln/eln/issues/286


yselkowitz (approved):
Thank you for doing this @hroncok and for being the first SST to be converted.

PR #1475, opened 2025-08-06
Add java-25-openjdk to CentOS/RHEL 10

gnu-andrew:
See https://issues.redhat.com/browse/RHEL-100678

### DIFF
--- a/configs/filelist
+++ b/configs/filelist
@@ -1,0 +1,1 @@
+sst_java-java.yaml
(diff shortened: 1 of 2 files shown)


gnu-andrew:
> I think it would make more sense to have one config per version (openjdk21 and openjdk25), which would turn this into a one-liner.

Can you give an example? I'm not sure how this become a one-liner when we need to add a whole bunch of packages.

> While we're at it, what's your new team name?

We don't have one as far as I'm aware.

yselkowitz:
See #1480

PR #1479, opened 2025-08-13
Application whitelisting is maintained by rhel-sst-security-selinux now

bachradsusi:


yselkowitz (changes requested):
`-sst` is no longer part of team names, can you please change this to the new team name?

yselkowitz (approved):

PR #1480, opened 2025-08-13
Reorganize Java workloads, add Java 25 to c10s

yselkowitz:
An alternative to #1475.

/cc @gnu-andrew


gnu-andrew:
I'm ok with this if you don't mind `eln` sharing with `c10s` again. They were split apart in this change you made - https://github.com/minimization/content-resolver-input/pull/1451 - and I didn't want to change stuff that was nothing to do with me.

PR #1481, opened 2025-08-14
Remove HA image workload from ELN. It should be CS only.

tdawson:


yselkowitz:
I know this was my idea, but then I saw other issues with HA packages come up (see https://github.com/fedora-eln/eln/issues/229), so I'm trying to convert these to use ELN branches and then we can treat them normally.  So don't merge this quite yet.

yselkowitz:
This will soon be fixed.

PR #1486, opened 2025-08-20
Make HighAvailability fully functional in ELN

yselkowitz:
Also, update ICU version


yselkowitz:
Waiting to merge until there is a compose with the ELN-branched fence-agents and pcs.

PR #1493, opened 2025-09-05
Update pcp subpackages in ELN

yselkowitz:
https://src.fedoraproject.org/rpms/pcp/c/9515e8a51fba8ceb02438afc72985d239ce91c7b

/cc @natoscott


yselkowitz:
Thanks for clarifying, I delisted `chan` and `HdrHistogram_c` as discussed above, they'll stay in ELN as dependencies as long as pcp requires them.

PR #1499, opened 2025-09-26
add virt-firmware-rs to sst-virt-firmware

kraxel:


kraxel:
Yes, the plan is to add it to rhel, PRP is here: https://issues.redhat.com/browse/RHELMISC-17815

yselkowitz (approved):

PR #1500, opened 2025-09-29
Add rhc-playbook-verifier as a dependency for rhc-worker-playbook

pkoprda:
* Resolves: RHELMISC-18048

Since rhc-playbook-verifier will be a new RPM for RHEL 10.2, this package should be a dependency of rhc-worker-playbook.


yselkowitz (changes requested):
This needs to be declared as a placeholder in its own right, and the dependency ignored (since CR doesn't handle that well).

yselkowitz (changes requested):

yselkowitz (approved):

PR #1520, opened 2025-11-13
add samestack to rts sst perf group

wchwhite:
This is for a prp process to add samestack to rts perf tools.


yselkowitz:
@wchwhite reopen or file a new ticket when you are ready to proceed with adding this to Fedora and RHEL.

PR #1586, opened 2026-04-15
Update KDE Gear packages to 26.04

yselkowitz:
- **Update KDE Gear packages for 26.04**
- **Add annobin clang/llvm plugins**

/cc @tdawson


tdawson:
Looks good.  Thank you for keeping up with this.

PR #1592, opened 2026-04-24
Drop unused libkolabxml

yselkowitz:
The Kolab resource in kdepim-runtime was removed in 26.04.0.

https://invent.kde.org/pim/kdepim-runtime/-/merge_requests/154

/cc @tdawson


tdawson:
Good to know.  That dependency always tripped me up.

PR #1601, opened 2026-05-20
Update langpacks-ii, delist jitterentropy

yselkowitz:
- **langpacks ii support is default fonts only**
- **Delist jitterentropy**


yselkowitz:
To be clear, we do still want jitterentropy, but we avoid listing dependencies so that if they become unused, they drop off and draw the appropriate scrutiny.

PR #1602, opened 2026-05-28
Remove fdo-rs packages from edge workload

djach7:
The Rust-based FIDO Device Onboard (fdo-rs) implementation is being removed in RHEL 11. The Go-based implementation (go-fdo-*) is the supported alternative.

https://redhat.atlassian.net/browse/THEEDGE-4607

### DIFF
--- a/configs/rhel-sst-edge--all.yaml
+++ b/configs/rhel-sst-edge--all.yaml
@@ -24,12 +24,6 @@ data:
   arch_packages:
     aarch64:
       - clevis-pin-tpm2
-      - fdo-client
-      - fdo-init
-      - fdo-manufacturing-server
-      - fdo-owner-cli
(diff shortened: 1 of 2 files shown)


djach7:
Hi @yselkowitz, are you able to take a look at this one? Thank you!

yselkowitz (changes requested):
@djach7 thanks for the PR.  Because these packages must remain (albeit deprecated) in RHEL 10, they cannot be delisted completely.  You will need to make a new config file with just these packages for only the c10s config to cover that.

djach7:
@yselkowitz Gotcha, thanks for the explanation. I just added a config file to this pr that contains just the deprecated packages, is there anything else I need to change?

yselkowitz (approved):

PR #1606, opened 2026-06-11
Remove weldr-client from eln

ondrejbudai:
weldr-client is deprecated in RHEL 10 and scheduled for removal in RHEL 11, so let's remove it from eln.


yselkowitz (approved):
This is fine by itself, but `osbuild-composer-tests` still depends on it (via the `composer-cli` alias), so that dependency will need to be dropped in order to remove it completely; see:

https://github.com/osbuild/osbuild-composer/blob/main/osbuild-composer.spec#L382
https://github.com/osbuild/osbuild-composer/pull/5096

ondrejbudai:
Let's drop it as the first step to remove the old image builder stack: https://github.com/osbuild/osbuild-composer/pull/5179

PR #1607, opened 2026-06-15
Update libzpc packages for ELN

yselkowitz:
As of libzpc 2.0.0, the shared library has been dropped, an OpenSSL provider added for s390x, and a multi-arch tool added:

https://src.fedoraproject.org/rpms/libzpc/c/8e47575a8a96dd97949e66a048a9246e464064b5

@sharkcz can you please confirm the following:

* That the standalone library is no longer wanted for RHEL 11?
* If the tool should be shipped on all arches?
* The manpages require `pandoc` which is not included in RHEL, so the RHEL package will manage without them?
* Will this change be coming back to RHEL 10?

### DIFF
--- a/configs/rhel-sst-arch-hw--c10s.yaml
+++ b/configs/rhel-sst-arch-hw--c10s.yaml
@@ -24,7 +24,6 @@ data:
     - alsa-utils
     - speexdsp
     - hwloc
-    - libzpc-tools
     - pigz
     - s390utils
     - s390utils-se-data
@@ -49,7 +48,8 @@ data:
       - libzfcphbaapi
       - libzfcphbaapi-devel
       - libzfcphbaapi-docs
-      - libzpc-provider
+      - libzpc
+      - libzpc-devel
       - openssl-ibmca
       - qclib
       - qclib-devel
@@ -75,4 +75,4 @@ data:
       - pcm
       - tboot
   labels:
-    - eln
+    - c10s


sharkcz:
> As of libzpc 2.0.0, the shared library has been dropped, an OpenSSL provider added for s390x, and a multi-arch tool added:
> 
> https://src.fedoraproject.org/rpms/libzpc/c/8e47575a8a96dd97949e66a048a9246e464064b5
> 
> @sharkcz can you please confirm the following:
> 
>     * That the standalone library is no longer wanted for RHEL 11?

yes, that's what we agreed on with IBM
 
>     * If the tool should be shipped on all arches?

yes
 
>     * The manpages require `pandoc` which is not included in RHEL, so the RHEL package will manage without them?

we had this in `s390utils` in the past and the fix was to provide prebuilt manpages in the source archive, plus some logic in project's buildsystem (IIRC)
 
>     * Will this change be coming back to RHEL 10?

for RHEL 10 the plan is to introduce `libzpc2` and have them both available so we don't cause API/dependency breakage during RHEL 10 lifetime. It will need some `libzpc2.el10` -> `libzpc.el11` upgrade path "magic".

yselkowitz:
Corresponding compose changes:
https://forge.fedoraproject.org/releng/pungi-fedora/pulls/1658
https://forge.fedoraproject.org/releng/fedora-comps/pulls/1287

PR #1608, opened 2026-06-15
Temporarily disable python3-testslide

yselkowitz:
This has yet to be rebuilt for Python 3.15, and is breaking the rest of the payload as a result.


yselkowitz:
fyi @michel-slm

PR #1611, opened 2026-06-19
Drop yggdrasil-worker-package-manager from ELN

yselkowitz:
https://redhat.atlassian.net/browse/RHEL-143958


rehana-raj:
As PO who is responsible for managing yggdrasil-worker-package-manager, I agree that this package should be removed from RHEL 11 as it is no longer relevant. https://redhat.atlassian.net/browse/RHEL-143958

PR #1616, opened 2026-06-25
Bulk rename for four Core Services teams,

notroj:
rhel-sst-cs-infra-services*… still to do.


yselkowitz (approved):
THANK YOU!!!  I had *not* been looking forward to doing this myself.

PR #1619, opened 2026-07-08
OpenSSL: Add fips-provider

### LINKED ISSUE #578: Add fips-provider(-next) to ELN
https://src.fedoraproject.org/rpms/openssl/c/b48db9c1ece416bd58eade1fff98eb00f4338ccf adds a dependency on a separate FIPS provider virtual package, which is provided by the new `fips-provider` package in Fedora, corresponding to `fips-provider-next` in RHEL.  We've decided to use the renaming dance for ELN.

This is needed ASAP to fix ELN composes.

sgallagher:
Splits the c10s and ELN configs into separate files, since c10s does not have fips-provider.

### DIFF
--- a/configs/rhel-sst-security-crypto--openssl.yaml
+++ b/configs/rhel-sst-security-crypto--openssl.yaml
@@ -8,6 +8,7 @@ data:
   maintainer: rhel-sst-security-crypto
   packages:
     - openssl
+    - fips-provider-next
   package_placeholders:
     - srpm_name: openssl-fips-provider
       rpms:
@@ -18,11 +19,6 @@ data:
           description: FIPS module for OpenSSL
           dependencies:
             - coreutils
-    - srpm_name: fips-provider-next
-      rpms:
-        - rpm_name: fips-provider-next
-          description: Next FIPS module, before it is validated
-          dependencies: []
   labels:
     - eln
     - c10s


yselkowitz (changes requested):
We're still discussing what this should look like, but whatever it is will probably be different.

configs/rhel-sst-security-crypto--openssl.yaml
```diff
@@ -8,6 +8,7 @@ data:
   maintainer: rhel-sst-security-crypto
   packages:
     - openssl
+    - fips-provider
```
yselkowitz: Based on https://github.com/fedora-eln/eln/issues/578, this should be:
```suggestion
    - fips-provider-next
```

configs/rhel-sst-security-crypto--openssl.yaml
```diff
@@ -25,4 +26,3 @@ data:
           dependencies: []
```
yselkowitz: This placeholder can then be removed, and no need for the split configs.

yselkowitz (changes requested):

sgallagher:
@yselkowitz I think this is correct now, based on the creation of the `fips-provider-next` [package](https://koji.fedoraproject.org/koji/packageinfo?packageID=43770) in Fedora ELN.

yselkowitz (approved):

PR #1625, opened 2026-07-14
Add rhel-sst-security-crypto--crypto-auditing workload

ueno:


yselkowitz (approved):

configs/rhel-sst-security-crypto--crypto-auditing.yaml
```diff
@@ -0,0 +1,11 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: Trace and monitor cryptographically-relevant events on the system
```
yselkowitz: name should be a short title

yselkowitz (changes requested):

configs/rhel-sst-security-crypto--crypto-auditing.yaml
ueno: Changed it to "crypto-auditing services and utilities"

yselkowitz (approved):

PR #1631, opened 2026-07-20
edge: add oci-delta

djach7:
Add oci-delta package to the edge workload and include c9s label for RHEL 9 support.

### DIFF
--- a/configs/rhel-sst-edge--all.yaml
+++ b/configs/rhel-sst-edge--all.yaml
@@ -19,6 +19,7 @@ data:
     - ignition
     - ignition-edge
     - nss-altfiles
+    - oci-delta
     - ostree
     - rpm-ostree
   arch_packages:


configs/rhel-sst-edge--all.yaml
```diff
@@ -45,4 +46,5 @@ data:
       - greenboot-default-health-checks
   labels:
     - eln
+    - c9s
```
yselkowitz: `c9s` is no longer tracked in Content Resolver, so this tag should not be added here.  (That does NOT prevent its addition to RHEL 9.)

yselkowitz (changes requested):

configs/rhel-sst-edge--all.yaml
djach7: Just updated, thank you

yselkowitz (approved):

PR #1637, opened 2026-07-28
Add .NET 11 (dotnet11.0) to c10s

omajid:
It's not packaged in Fedora yet.

We are adding it to c10s but given the short EOL, probably not to c11s/eln.

### DIFF
--- a/configs/rhel-pt-dotnet--dotnet11.yaml
+++ b/configs/rhel-pt-dotnet--dotnet11.yaml
@@ -1,0 +1,14 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: .NET 11
+  description: .NET 11 SDK, tools and runtime
+  maintainer: rhel-pt-dotnet
+  packages:
+    # EOL in Nov, 2028
+    - dotnet-sdk-aot-11.0
+    - dotnet-sdk-11.0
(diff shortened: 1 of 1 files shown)


yselkowitz:
@omajid actually I wanted to ask you about this.  With the (relatively) recent change to STS support cycles, .NET 11 EOL should be the same as .NET 10, November 2028, in which case I was wondering if it would make more sense to switch ELN to dotnet11.0 once it is in rawhide (and drop 10.0), and then use (only) that for the RHEL 11 branching.  Then, presumably, you would update RHEL 11 to .NET 12 prior to GA.

yselkowitz (approved):
Either way, since this isn't in rawhide yet, we can't add it to ELN yet either.

omajid:
@yselkowitz Good questions. Let me talk to others in my team and get back to you.

omajid:
@yselkowitz my team and I talked about this and your suggestion is very sensible. We will do that.

yselkowitz:
Very well, do you have an estimated target for landing dotnet11.0 in rawhide?

omajid:
I am almost ready to propose the package for Fedora, then it's up to the review process and how long it takes to build/bootstrap it. Hopefully should be ready in a month or two.

PR #1651, opened 2026-08-06
ELN Extras: Include ansible-packaging

sgallagher:
Required for any package that uses the `%{ansible_collection_url}` macro.

### DIFF
--- a/configs/epel-gotmax23--all.yaml
+++ b/configs/epel-gotmax23--all.yaml
@@ -8,5 +8,6 @@ data:
     - ansible-collection-ansible-posix
     - ansible-collection-community-crypto
     - ansible-collection-community-general
+    - ansible-packaging
   labels:
     - eln-extras


sgallagher:
This will fix the failure to build `ansible-collection-community-general`. I suspect several others are "succeeding" because the `%{ansible_collection_url}` macro without arguments just resolves to the empty string, but this package uses arguments which break when the actual macro definition is not present.

yselkowitz:
> This will fix the failure to build `ansible-collection-community-general`. I suspect several others are "succeeding" because the `%{ansible_collection_url}` macro without arguments just resolves to the empty string, but this package uses arguments which break when the actual macro definition is not present.

No, it won't.  This works in EPEL because [epel-rpm-macros Requires: ansible-srpm-macros](https://src.fedoraproject.org/rpms/epel-rpm-macros/c/eb6d5b5bc15390f98d7fd437bdae2b70e58a1d7d?branch=epel10) which provides the macro.  Without something to pull it into the *default* buildroot, the buildSRPMFromSCM will fail, and there will never be a build.

yselkowitz:
IOW, getting these to build in Extras is blocked on https://github.com/fedora-eln/eln/issues/187 which probably requires https://github.com/fedora-eln/elnbuildsync/issues/60, and getting this listed as a BR is blocked on https://github.com/fedora-eln/content-resolver/issues/72

yselkowitz:
epel-rpm-macros is now in Extras which pulls this in, but making the dependent packages buildable requires work elsewhere.

PR #1656, opened 2026-08-16
kde: add kdominate, keepsecret to Gear

yselkowitz:
These are new as of 26.08:

https://community.kde.org/KDE_Gear/26.08_Release_notes

/cc @tdawson


tdawson:
Thank you for the heads up.

PR #1658, opened 2026-08-19
Drop audiotube from Extras

yselkowitz:
This requires yt-dlp which was in EPEL 9 at one point but subsequently dropped:

https://lists.fedoraproject.org/archives/list/epel-devel@lists.fedoraproject.org/message/6SWXFQP43JDZOHAMGOCEGBB22OQZO4X7/

See also: https://github.com/fedora-eln/eln/issues/606

/cc @tdawson


tdawson:
audiotube never made it into epel9 or epel10, so I'm ok with this.

PR #1660, opened 2026-09-04
cloud: Include cvmutils

vittyvk:
cvmutils package is being used to encrypt RHEL CVM OS images in Azure.

Signed-off-by: Vitaly Kuznetsov <vkuznets@redhat.com>

### DIFF
--- a/configs/rhel-sst-virtualization-cloud--cloud_azure_cvm.yaml
+++ b/configs/rhel-sst-virtualization-cloud--cloud_azure_cvm.yaml
@@ -1,0 +1,16 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: RHEL on Microsoft Azure Confidential VMs
+  description: Tools to assist running RHEL as a guest on Microsoft Azure CVMs
+  maintainer: rhel-sst-virtualization-cloud
+  packages: []
+  arch_packages:
+    aarch64:
+      - cvm-encrypt-image
+      - cvm-reseal
+    x86_64:
+      - cvm-encrypt-image
+      - cvm-reseal
+  labels:
+    - eln


configs/rhel-sst-virtualization-cloud--cloud_azure.yaml
```diff
@@ -7,6 +7,9 @@ data:
   packages:
     - cloud-init
     - cloud-utils-growpart
+    - cvm-encrypt-image
+    - cvm-reseal
+    - python3-cvmutils
```
yselkowitz: Dependencies should not be listed, so drop python3-cvmutils.

yselkowitz (changes requested):
Is this also going into RHEL 10?  I can't find a PRP for this.

vittyvk:
Not pushing it to RHEL10 yet but we certainly want it in RHEL11. The package has landed in Fedora recently.

yselkowitz:
> Not pushing it to RHEL10 yet but we certainly want it in RHEL11. The package has landed in Fedora recently.

Note that this config is also tagged `c10s`, so in this case you should create a new Azure CVM config with just these packages, and tagged only `eln`.

yselkowitz:
Presumably this will only be shipped on supported architectures (just x86_64 iiuc)?  If so, this should be treated here like an arch-specific package, even if the RPMs are technically noarch.

vittyvk:
> Presumably this will only be shipped on supported architectures (just x86_64 iiuc)? If so, this should be treated here like an arch-specific package, even if the RPMs are technically noarch.

Currently, Azure CVMs where this stuff is used is x86 only and the tool also supports x86 VM images exclusively. Going forward, I can see this being used on aarch64 instances as well with a few fairly trivial code changes which I believe will be made even before RHEL11 fork. There are no short or mid term plans to support other arches. I will update the PR accordinly.

yselkowitz (approved):

PR #1662, opened 2026-09-09
Update NVIDIA Extensions packages

mizdebsk:
Sync NVIDIA packages workload to match expected Extensions content.

Some of the packages may be unavailable in Fedora yet. The goal is to see what is missing and either package it, or convert to a placeholder. If that's not the intended use of the Extensions workloads then I will limit the workload to only packages already available in Fedora.


configs/rhel-extensions--nvidia.yaml
```diff
@@ -0,0 +1,21 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: NVIDIA packages
+  description: NVIDIA packages in the RHEL Extensions repository
+  maintainer: rhel-gpuaccelerators-nvidia-rm
+  packages:
+    - dnf-plugin-nvidia
```
yselkowitz: `dnf-plugin-nvidia` is not in Fedora, so it needs to be a placeholder.

configs/rhel-extensions--nvidia.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - egl-gbm
+    - egl-wayland
+    - egl-wayland-devel
+    - egl-wayland2
+    - egl-x11
+    - eglexternalplatform-devel
+    - nvidia-libXNVCtrl
+    - nvidia-libXNVCtrl-devel
```
yselkowitz, line 15: This is in Fedora as `libXNVCtrl` and Obsoletes/Provides the `nvidia`-prefixed names.  While technically either will work for now, the concern is if Fedora ever drops those Obsoletes/Provides (which ordinarily happens after a few Fedora releases, but may not make sense in this particular case).

configs/rhel-extensions--nvidia.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - nvidia-modprobe
+    - nvidia-persistenced
+    - nvidia-settings
```
yselkowitz: These are not in Fedora and therefore need to be placeholders.

yselkowitz (changes requested):

configs/rhel-extensions--nvidia.yaml
mizdebsk: Fixed

configs/rhel-extensions--nvidia.yaml
mizdebsk: Fixed

configs/rhel-extensions--nvidia.yaml
mizdebsk: I would prefer to use `nvidia-*` names to match how the actual Extensions packages are named like.

configs/rhel-extensions--nvidia.yaml
```diff
@@ -14,8 +13,74 @@ data:
...
+  package_placeholders:
+    - srpm_name: dnf-plugin-nvidia
+      build_dependencies:
+        - python3-devel
+      rpms:
+        - rpm_name: dnf-plugin-nvidia
+          description: DNF plugin needed to remove old kernel modules
+          dependencies:
+            - python3
+            - python3-dnf
```
yselkowitz: As far as ELN Extensions is trying to prepare for RHEL 11 Extensions (on the presumption that packages now in 10 Extensions may also be needed in 11, although that's not guaranteed), this will need to be replaced by a "(lib)dnf5-plugin-nvidia" which is compatible with dnf5, as that will replace dnf in RHEL 11.  I'm not sure if that even exists yet though?

yselkowitz (changes requested):

configs/rhel-extensions--nvidia.yaml
mizdebsk: DNF5 plugin does not yet exist and we are aware of that.
I am working under assumption that one of the main points of adding the workloads and placeholders was exactly to catch issues like this.

yselkowitz: Well, consider it caught.  But merging this would create a new dependency on `python3-dnf` and prevent it from dropping (which we're trying to do).  IOW we're not keeping dnf-4 in RHEL 11 just because this hasn't been ported yet.  Please be in touch with NVIDIA about the need to port this to libdnf5 (for both Fedora and RHEL 11), but in the meantime, either this needs to be adjusted to match what will be in the future, or omitted (commented out) until a solution for dnf5 exists.

configs/rhel-extensions--nvidia.yaml
```diff
@@ -0,0 +1,86 @@
+document: feedback-pipeline-workload
+version: 1
+data:
+  name: NVIDIA packages
+  description: NVIDIA packages in the RHEL Extensions repository
+  maintainer: rhel-gpuaccelerators-nvidia-rm
+  packages:
+    - egl-gbm
+    - egl-wayland
+    - egl-wayland-devel
```
yselkowitz, line 10: Another question, will `egl-wayland` still be needed now that there is `egl-wayland2`?

mizdebsk: For the time being yes, both are needed because different NVIDIA driver versions need different EGL packages. Eventually we'll be able to drop `egl-wayland` and keep only `egl-wayland2`.

configs/rhel-extensions--nvidia.yaml
mizdebsk: Should be fixed.

configs/rhel-extensions--nvidia.yaml
yselkowitz: Then perhaps this package should be renamed in Fedora?

yselkowitz (approved):

PR #1664, opened 2026-09-11
add qatlib-tools subpackage

nefigtut:
Fedora package: https://src.fedoraproject.org/rpms/qatlib
Jira: https://redhat.atlassian.net/browse/RHEL-258469


yselkowitz (changes requested):
qatlib-tools is ok, but jitterentropy is just a dependency, and we intentionally do not list dependencies here so that they can drop automatically when no longer required.

nefigtut:
thanks, Yaakov @yselkowitz, updated as you suggest, thank you. can you please re-review?

yselkowitz (approved):

PR #1667, opened 2026-09-14
Update dnf-plugin-nvidia placeholder for DNF 5

mizdebsk:
Follow up on #1662

The dnf-plugin-nvidia placeholder is updated to reflect major version 3 of the plugin that targets DNF 5.

Package source: https://gitlab.com/nvidia/packages/redhat/dnf-plugin-nvidia/-/tree/fc44?ref_type=heads


yselkowitz:
Shouldn't this be renamed to `dnf5-plugin-nvidia`?  Everything else uses the `dnf5-` prefix now.

mizdebsk:
> Shouldn't this be renamed to `dnf5-plugin-nvidia`? Everything else uses the `dnf5-` prefix now.

No, the upstream package name is still `dnf-plugin-nvidia` as you can see in the linked source repository.

yselkowitz:
The upstream name is irrelevant, what matters is how it will be packaged downstream in Fedora and/or RHEL.  Why wouldn't this be named `dnf5-plugin-nvidia`?

mizdebsk:
Upstream name is very relevant as we want to maintain compatibility with NVIDIA packaging.
NVIDIA proprietary packages (which we don't have control over) are and will be referring to this package by the name `dnf-plugin-nvidia`.

PR #1668, opened 2026-09-15
Drop subscription-manager-plugin-ostree from ELN

yselkowitz:
This package is deprecated: https://redhat.atlassian.net/browse/RHEL-95688


rehana-raj:
As a PO of subscription-manager-plugin-ostree, I confirm that this rpm is expected to removed from the future RHEL versions.

